PR #1, opened 2022-01-25
Implement queue data model

corneliouzbett:
- Initial iteration to implement service delivery queues, data model. This is based on the final decision/thoughts [here](https://talk.openmrs.org/t/backend-support-for-service-delivery-queues-workflows/35247/52)

service-delivery-queue-ER
.

### DIFF
--- a/api/src/main/java/org/openmrs/module/queue/model/EncounterStatus.java
+++ b/api/src/main/java/org/openmrs/module/queue/model/EncounterStatus.java
@@ -1,0 +1,76 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module.queue.model;
+
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.Table;
+import java.util.Date;
+
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+import lombok.NoArgsConstructor;
+import org.openmrs.BaseOpenmrsMetadata;
+import org.openmrs.Concept;
+import org.openmrs.Encounter;
+import org.openmrs.EncounterType;
+import org.openmrs.Patient;
+import org.openmrs.Visit;
+
+@EqualsAndHashCode(onlyExplicitlyIncluded = true, callSuper = true)
+@NoArgsConstructor
+@Data
+@Entity
+@Table(name = "encounter_status")
+public class EncounterStatus extends BaseOpenmrsMetadata {
+
+	private static final long serialVersionUID = 1L;
+
+	@Id
+	@EqualsAndHashCode.Include
+	@GeneratedValue(strategy = GenerationType.AUTO)
+	@Column(name = "encounter_status_id")
+	private Integer encounterStatusId;
+
+	@Column(name = "patient_id")
+	private Patient patient;
+
+	@Column(name = "encounter_type_id")
+	private EncounterType encounterType;
+
+	@Column(name = "encounter_id")
+	private Encounter encounter;
+
+	@Column(name = "encounter_datetime")
+	private Date encounterDatetime;
+
+	/**
+	 * Note: Visit_id will be used to help unambiguously identify the encounter to which the status belongs not meant
+	 * to assign status directly to a visit.
+	 */
+	@Column(name = "visit_id")
+	private Visit visit;
+
+	@Column(name = "status", nullable = false)
+	private Concept status;
+
+	@Override
+	public Integer getId() {
+		return getEncounterStatusId();
+	}
+
+	@Override
+	public void setId(Integer id) {
+		this.setEncounterStatusId(id);
+	}
+}
--- a/api/src/main/java/org/openmrs/module/queue/model/EncounterStatus.java
+++ b/api/src/main/java/org/openmrs/module/queue/model/EncounterStatus.java
@@ -27,7 +27,7 @@ import org.openmrs.EncounterType;
 import org.openmrs.Patient;
 import org.openmrs.Visit;
 
-@EqualsAndHashCode(onlyExplicitlyIncluded = true, callSuper = true)
+@EqualsAndHashCode(callSuper = true)
 @NoArgsConstructor
 @Data
 @Entity
@@ -37,7 +37,6 @@ public class EncounterStatus extends BaseOpenmrsMetadata {
 	private static final long serialVersionUID = 1L;
 
 	@Id
-	@EqualsAndHashCode.Include
 	@GeneratedValue(strategy = GenerationType.AUTO)
 	@Column(name = "encounter_status_id")
 	private Integer encounterStatusId;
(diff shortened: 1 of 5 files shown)


api/src/main/java/org/openmrs/module/queue/model/EncounterStatus.java
```diff
@@ -0,0 +1,76 @@
...
+@NoArgsConstructor
+@Data
+@Entity
+@Table(name = "encounter_status")
+public class EncounterStatus extends BaseOpenmrsMetadata {
+
+	private static final long serialVersionUID = 1L;
+
+	@Id
+	@EqualsAndHashCode.Include
```
ibacher, line 40: Usually, we use the UUID as the equality check for domains. Is there a reason to do otherwise here?

api/src/main/java/org/openmrs/module/queue/model/QueueEntry.java
```diff
@@ -0,0 +1,81 @@
...
+	@Column(name = "priority", nullable = false)
+	private Concept priority;
+
+	@Column(name = "priority_comment")
+	private String priorityComment;
+
+	@Column(nullable = false)
+	private Concept status;
+
+	private double sort_weight;
```
ibacher, line 56:
```suggestion
	@Column(name = "sort_weight")
	private double sortWeight;
```

api/src/main/java/org/openmrs/module/queue/model/EncounterStatus.java
```diff
@@ -0,0 +1,76 @@
...
+import org.openmrs.EncounterType;
+import org.openmrs.Patient;
+import org.openmrs.Visit;
+
+@EqualsAndHashCode(onlyExplicitlyIncluded = true, callSuper = true)
+@NoArgsConstructor
+@Data
+@Entity
+@Table(name = "encounter_status")
+public class EncounterStatus extends BaseOpenmrsMetadata {
```
ibacher, line 35: EncounterStatus, like Encounter, is really data rather than metadata (that is to say, we want the semantics of "voided" rather than "retired").

```suggestion
public class EncounterStatus extends BaseChangeableOpenmrsData {
```

api/src/main/java/org/openmrs/module/queue/model/QueueEntry.java
```diff
@@ -0,0 +1,81 @@
...
+import org.openmrs.Location;
+import org.openmrs.Patient;
+import org.openmrs.Provider;
+
+@EqualsAndHashCode(callSuper = true)
+@NoArgsConstructor
+@Data
+@Entity
+@Table(name = "queue_entry")
+public class QueueEntry extends BaseOpenmrsMetadata {
```
ibacher, line 35:
```suggestion
public class QueueEntry extends BaseChangeableOpenmrsData {
```

api/src/main/java/org/openmrs/module/queue/model/EncounterStatus.java
ibacher: I guess `BaseChangeableOpenmrsData` is only available in 2.2+

ibacher (commented):
Thanks @corneliouzbett! Couple of small things. I think @bmamlin had also wanted a Queue table in there (which would be proper metadata) to capture queue names and descriptions.

api/src/main/java/org/openmrs/module/queue/model/EncounterStatus.java
corneliouzbett: I've always treated "retired" & "voided" as the same thing. Now I know, Thanks!

corneliouzbett:
I missed the part on `queue` for metadata and `queue_stats` which I haven't figured out yet(what kind of stats on queues do O3 wants). 
@ibacher  At what point do we stop supporting certain versions of openmrs platform i.e < `2.0` or even < `2.2`?

ibacher:
Let's ignore `queue_stats` for the moment and just keep it as a potential concept... I think we're fine with targetting 2.3.x as the minimal version at least for now. That should cover OHRI, AMPATH, and PIH.

api/src/main/resources/liquibase.xml
```diff
@@ -1,8 +1,197 @@
...
+                <constraints nullable="false"/>
+            </column>
+            <column name="changed_by" type="int"/>
+            <column name="date_changed" type="datetime"/>
+            <column name="retired" type="boolean" defaultValueBoolean="false">
+                <constraints nullable="false"/>
+            </column>
+            <column name="retired_by" type="int"/>
+            <column name="date_retired" type="datetime"/>
+            <column name="retire_reason" type="varchar(255)" defaultValue="null"/>
```
dkayiwa, line 41: I think the default value of `null` was a typo that got carried over in many places. 😊

api/src/main/resources/liquibase.xml
```diff
@@ -1,8 +1,197 @@
...
+                <constraints nullable="false"/>
+            </column>
+            <column name="changed_by" type="int"/>
+            <column name="date_changed" type="datetime"/>
+            <column name="voided" type="boolean" defaultValueBoolean="false">
+                <constraints nullable="false"/>
+            </column>
+            <column name="voided_by" type="int"/>
+            <column name="date_voided" type="datetime"/>
+            <column name="void_reason" type="varchar(255)" defaultValue="null"/>
```
dkayiwa, line 100: Same as above for the default value of `null`.

api/src/main/resources/liquibase.xml
```diff
@@ -1,8 +1,197 @@
...
```
dkayiwa, line 166: Same as above.

api/src/main/java/org/openmrs/module/queue/model/EncounterStatus.java
```diff
@@ -0,0 +1,75 @@
...
+	@Column(name = "encounter_status_id")
+	private Integer encounterStatusId;
+
+	@Column(name = "patient_id")
+	private Patient patient;
+
+	@Column(name = "encounter_type_id")
+	private EncounterType encounterType;
+
+	@Column(name = "encounter_id")
```
dkayiwa, line 50: Is there a possibility of this encounter having a different encounter type and encounter date time from those in the `encounterType` and `encounterDatetime` fields?

ibacher: @dkayiwa I think the idea here is that an encounter status might exist before an actual encounter has started. So we can use the `encounterType` field to track what sort of encounter *should* be created. Then, once the encounter actually exists, we reference the encounter for that information.

dkayiwa: Oh i see! Thanks for the explanation. 😊

api/src/main/resources/liquibase.xml
corneliouzbett: Yeah, that must be it.

api/src/main/resources/liquibase.xml
```diff
@@ -180,15 +180,6 @@
         <addForeignKeyConstraint baseTableName="encounter_status" baseColumnNames="patient_id"
                                  constraintName="encounter_status_patient_id_fk"
                                  referencedTableName="patient" referencedColumnNames="patient_id"/>
-        <addForeignKeyConstraint baseTableName="encounter_status" baseColumnNames="encounter_type_id"
-                                 constraintName="encounter_status_encounter_type_id_fk"
-                                 referencedTableName="encounter_type" referencedColumnNames="encounter_type_id"/>
-        <addForeignKeyConstraint baseTableName="encounter_status" baseColumnNames="encounter_id"
```
ibacher: Shouldn't we at least have the FKs on encounter_type and encounter?

corneliouzbett: Yes. My initial thinking was to loosen the restrictions when creating `encounter_status`, without necessarily starting a visit.

ibacher: I think making sure the columns are nullable should be fine. The FKs are semi-important as they create indexes that can be used for joins, i.e., SQL will be somewhat more efficient.

ibacher (approved):

PR #2, opened 2022-01-31
Add encounter_status queue_entry mapping table

corneliouzbett:
- A junction table between `encounter_status` & `queue_entry`


corneliouzbett:
To be re-open when the need arise :)

PR #3, opened 2022-02-04
Feat: Initial queue mvp

corneliouzbett:
Well, this PR tries to do sort of "everything". Initial implementation of generic queue MVP; below is an updated image of the data model.

queue ER

Pending todo items;
- Define query entries list - How to get an ordered list of queue entries; 
- Visits + queue + queue-entry integration - when to create a `visit-query-entries` record
- Calculate queue average time
- Return queue count


api/src/main/java/org/openmrs/module/queue/model/VisitQueueEntry.java
```diff
@@ -0,0 +1,88 @@
...
+import org.hibernate.search.annotations.Field;
+import org.openmrs.Auditable;
+import org.openmrs.Retireable;
+import org.openmrs.User;
+import org.openmrs.Visit;
+
+@Data
+@Entity
+@Table(name = "visit_queue_entries")
+public class VisitQueueEntry implements Auditable, Retireable {
```
ibacher, line 36: Could this just be "extends BaseOpenmrsData"?

ibacher (approved):
Awesome! Thanks @corneliouzbett!

PR #4, opened 2022-02-11
Validations, visit queue entry, and search implementations

corneliouzbett:
This PR wraps up CRUD operation implementation for the 3 resources and subresources, additionally, it adds basic(commons) validations rule which includes;
- `queueEntry.status` concept should be a member of the queue status concept set configured via GP; this applies to queue priority and queue service
- `queue.name` - not null
- `queue.location` - should exists, in future only location tag with queue(specific use case) 
- ...

Search Operations
-  Implement search by queue status - `/queue/<parentUuid>/entry?status=Waiting For Service`
- Implement count by status - `queue/<UUID>/count?status=Waiting For Service`
- ...

At this point, I think we got a working MVP of the queue module. Maybe one or two logic/features are missing. Ready for testing/usage.

### DIFF
--- a/omod/src/main/java/org/openmrs/module/queue/web/resources/QueueEntryCountSubResource.java
+++ b/omod/src/main/java/org/openmrs/module/queue/web/resources/QueueEntryCountSubResource.java
@@ -9,10 +9,12 @@
  */
 package org.openmrs.module.queue.web.resources;
 
-import java.util.Collections;
+import java.util.Arrays;
 
 import org.openmrs.module.queue.model.Queue;
-import org.openmrs.module.queue.web.resources.response.QueueCount;
+import org.openmrs.module.queue.model.QueueEntry;
+import org.openmrs.module.queue.web.resources.custom.response.GenericSingleObjectResult;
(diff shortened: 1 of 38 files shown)


omod/src/main/java/org/openmrs/module/queue/web/resources/QueueEntryCountSubResource.java
```diff
@@ -37,48 +38,48 @@
...
 	
 	@Override
-	public void setParent(QueueCount queueCount, Queue queue) {
-		queueCount.setQueue(queue);
+	public void setParent(QueueEntry queueEntryCount, Queue queue) {
+		queueEntryCount.setQueue(queue);
 	}
 	
 	@Override
 	public PageableResult doGetAll(Queue queue, RequestContext requestContext) throws ResponseException {
```
ibacher, line 56: Hadn't really clocked this before... It seems like this is a weird thing to do in `doGetAll()`, but I suppose if it works....

ibacher (approved):
LGTM

PR #16, opened 2023-01-10
03-1741 Add clinics to the queue model

CynthiaKamau:
- Added queue clinic model to the queue. Each queue belongs to a specific clinic/sub location in the hospital . This will enable distinguishing of similar service points in the same location. eg Clinical consultation in HIV Clinic and Clinical consultation in MCH clinic.

### DIFF
--- a/api/src/main/java/org/openmrs/module/queue/model/QueueClinic.java
+++ b/api/src/main/java/org/openmrs/module/queue/model/QueueClinic.java
@@ -16,8 +16,6 @@ import javax.persistence.FetchType;
 import javax.persistence.GeneratedValue;
 import javax.persistence.GenerationType;
 import javax.persistence.Id;
-import javax.persistence.JoinColumn;
-import javax.persistence.ManyToOne;
 import javax.persistence.OneToMany;
 import javax.persistence.Table;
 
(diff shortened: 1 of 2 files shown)


CynthiaKamau:
Please review @corneliouzbett @slubwama

corneliouzbett (commented):
I think treating the HIV clinic and MCH clinic as different locations but sharing the same parent location can be used to support this use case. Can you shed more light on why we need the new data model for this?

slubwama:
> * Added queue clinic model to the queue. Each queue belongs to a specific clinic/sub location in the hospital . This will enable distinguishing of similar service points in the same location. eg Clinical consultation in HIV Clinic and Clinical consultation in MCH clinic.

@CynthiaKamau  Why don't we use locations as we had discussed?

CynthiaKamau:
> > * Added queue clinic model to the queue. Each queue belongs to a specific clinic/sub location in the hospital . This will enable distinguishing of similar service points in the same location. eg Clinical consultation in HIV Clinic and Clinical consultation in MCH clinic.
> 
> @CynthiaKamau Why don't we use locations as we had discussed?

Because some implementations use multitenancy, so there has to be a way to distinguish the different facilities.

PR #30, opened 2023-04-20
Filter out past visit queue entries

makombe:


CynthiaKamau:
Kindly fix the failing tests

makombe:
@CynthiaKamau I think it is not really the tests but the maven build is what is failing. Thanks

PR #38, opened 2023-08-09
(feat) Add location coming from in queue entry model

CynthiaKamau:


api/src/main/java/org/openmrs/module/queue/model/QueueEntry.java
```diff
@@ -80,6 +80,11 @@ public class QueueEntry extends BaseChangeableOpenmrsData {
 	@JoinColumn(name = "provider_waiting_for", referencedColumnName = "provider_id")
 	private Provider providerWaitingFor;
 	
+	//The Location the patient is coming for, if any.
```
ojwanganto, line 83: the comment should read location patient is coming for. just a small type

ojwanganto (approved):

api/src/main/resources/liquibase.xml
```diff
@@ -328,4 +328,25 @@
...
+            <column name="location_coming_from" type="int">
+            </column>
+        </addColumn>
+    </changeSet>
+
+    <changeSet id="queue_entry_20230809-02" author="cynthiakamau">
+        <preConditions onFail="MARK_RAN">
+            <columnExists tableName="queue_entry" columnName="location_coming_from"/>
+        </preConditions>
+        <addForeignKeyConstraint baseColumnNames="queue_entry" baseTableName="location_coming_from" constraintName="queue_entry_location_coming_from_id_fk" deferrable="false" initiallyDeferred="false" onDelete="NO ACTION" onUpdate="NO ACTION" referencedColumnNames="queue_id" referencedTableName="queue"/>
```
corneliouzbett, line 350: In the addForeignKeyConstraint statement, there are a couple of issues I see:
- The baseColumnNames should be `location_coming_from` not `queue_entry`. This specifies the column in the queue_entry table that the foreign key refers to.
- The baseTableName should be `queue_entry` not `location_coming_from`. This specifies the table that contains the column the foreign key constraint is on.

corneliouzbett: Both `deferrable="false"` and `initiallyDeferred="false"` are redundant since default values are false

api/src/main/resources/liquibase.xml
```diff
@@ -328,4 +328,25 @@
...
+        <comment>
+            Add column location_coming_from to queue entry table
+        </comment>
+        <addColumn tableName="queue_entry">
+            <column name="location_coming_from" type="int">
+            </column>
+        </addColumn>
+    </changeSet>
+
+    <changeSet id="queue_entry_20230809-02" author="cynthiakamau">
```
corneliouzbett, line 346:
```suggestion
    <changeSet id="add_queue_entry_location_coming_from_fk_20230809" author="cynthiakamau">
```

api/src/main/resources/liquibase.xml
```diff
@@ -328,4 +328,25 @@
             TIMESTAMP(CURDATE()), 'MM/dd/yyyy HH:mm:ss', 86400, TRUE, 1, NOW(), UUID());
         </sql>
     </changeSet>
+
+    <changeSet id="queue_entry_20230809-01" author="cynthiakamau">
```
corneliouzbett, line 332:
```suggestion
    <changeSet id="add_location_coming_from_column_to_queue_entry_20230809" author="cynthiakamau">
```

api/src/main/java/org/openmrs/module/queue/model/QueueEntry.java
```diff
@@ -80,6 +80,11 @@ public class QueueEntry extends BaseChangeableOpenmrsData {
...
+	@ManyToOne
```
corneliouzbett, line 84: Why `@ManyToOne`?

corneliouzbett (changes requested):

api/src/main/java/org/openmrs/module/queue/model/QueueEntry.java
CynthiaKamau: multiple instances of location_coming_from can be associated with a single queue

api/src/main/resources/liquibase.xml
CynthiaKamau: Done

api/src/main/resources/liquibase.xml
ojwanganto: Makes sense. Sorry I didn't pay much attention

api/src/main/java/org/openmrs/module/queue/model/QueueEntry.java
ojwanganto: It can be confusing whether to use one/many to one relationship. @corneliouzbett - I wonder what this will look like for those who are sent to the lab but has to go through billing. Based on the current design, what are your thoughts?

corneliouzbett: `@ManyToOne` is okay for this case.
However, I'm curious if the next queue is determined based on the client's origin.

corneliouzbett: Based on the current design, the clinical can send the client to billing and then from billing to the Lab (Clinical consultation -> Billing -> Lab -> Clinical Consultation (again)).
I suggest enabling clinical officers to control client movement. Once one stage is done, the client can be automatically placed in the next queue with a designated priority.

CynthiaKamau: > `@ManyToOne` is okay for this case. However, I'm curious if the next queue is determined based on the client's origin.

This is dependent on the clinic workflow. For now, the providers chose the queue that they want to move a patient to, its not done automatically yet

makombe: I concur with @corneliouzbett . With the current design we have, the patient will only have one active queue at a time. One can not belong to two different active queues. Once a patient is moved from one queue to another, the previous queue is ended.

CynthiaKamau: Done

makombe (approved):
LGTM! Thanks @CynthiaKamau

api/src/main/java/org/openmrs/module/queue/model/QueueEntry.java
```diff
@@ -80,6 +81,11 @@ public class QueueEntry extends BaseChangeableOpenmrsData {
 	@JoinColumn(name = "provider_waiting_for", referencedColumnName = "provider_id")
 	private Provider providerWaitingFor;
 	
+	//The Location the patient is coming from, if any.
+	@OneToOne
+	@JoinColumn(name = "location_coming_from", referencedColumnName = "queue_id")
+	private Queue locationComingFrom;
```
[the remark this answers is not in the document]
ojwanganto: @mseaton there is a talk post on the design of the service queues. I agree with the naming convention for queueComingFrom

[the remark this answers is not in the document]
mseaton: @ojwanganto thanks.  Yes, I'm aware of the talk post, but I don't recall any discussion there of adding this property to the QueueEntry object.  Can you point me to that please?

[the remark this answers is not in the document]
ojwanganto: Sorry I misunderstood your question. It is like we only created the task in the Palladium jira but we can create a ticket in the OpenMRS and have these discussions there. We already resolved to have all tickets relating to the module tracked in OpenMRS jira and we shouldn't have anything similar going forward. Will that be OK with you?

PR #39, opened 2023-09-14
O3-2382 Add column to denote patient previous queue (rename column)

CynthiaKamau:


mseaton (approved):
Change looks good to me @CynthiaKamau .  I would love to learn more about how you are using this column / property in your queue workflow that necessitated adding this in general, but assuming you mean for this to be a Queue and not a Location object, I agree with the change to naming.

CynthiaKamau:
> Change looks good to me @CynthiaKamau . I would love to learn more about how you are using this column / property in your queue workflow that necessitated adding this in general, but assuming you mean for this to be a Queue and not a Location object, I agree with the change to naming.

The column helps to know where the patient should go next based on the services they have already received

PR #40, opened 2023-09-26
O3-2443 AutoCloseQueueEntryTask could result in ending queue entries prematurely.

mseaton:
…prematurely


api/src/main/java/org/openmrs/module/queue/tasks/AutoCloseVisitQueueEntryTask.java
```diff
@@ -0,0 +1,77 @@
...
+ * This iterates over all active VisitQueueEntries If the Visit associated with any of these has
+ * ended (non-null stopDatetime), then the QueueEntry associated with it is also ended at the same
+ * datetime as the Visit was stopped.
+ */
+@Slf4j
+public class AutoCloseVisitQueueEntryTask implements Runnable {
+	
+	private static boolean currentlyExecuting = false;
+	
+	@Override
```
mseaton, line 33: These are the crux of the changes.  Whereas the previous approach simply closed every active queue entry every 24 hours, this approach removes that for one that is more limited in it's scope - only automatically close active queue entries if they are associated with active visits.  Because the goal is to constantly check to see if there are any visits that are out of sync with their queue entries, this is executed every minute and set to ensure that if it happens to take a minute to execute (it shouldn't) that it doesn't run again in parallel.

api/src/main/java/org/openmrs/module/queue/tasks/QueueTaskExecutor.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+package org.openmrs.module.queue.tasks;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.scheduling.concurrent.ScheduledExecutorFactoryBean;
+import org.springframework.scheduling.concurrent.ScheduledExecutorTask;
+import org.springframework.stereotype.Component;
+
+/**
```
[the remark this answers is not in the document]
ojwanganto: Thanks for sharing this alternative approach. We have benefited so much from it without knowing much about the implementation.

api/src/main/java/org/openmrs/module/queue/tasks/AutoCloseVisitQueueEntryTask.java
ojwanganto: This will work well if all visits could be closed at the end of day or as appropriate. Our main challenge initially was about active visits that span several days before they are closed as they still have active queue entries. With proper patient check out, this approach will ensure near realtime queue clean up

ojwanganto (approved):
It looks good to me

mseaton:
> It looks good to me

Thanks @ojwanganto . Note, I also pushed a unit test in a follow-up commit.  Regarding the visit closing, yes - in our implementation (and I believe in the refapp generally, but I'm not 100% sure) there is already a scheduled task that runs (daily I believe) to close visits that are open based on a set of rules.  So this would complement that - once the visits are closed, this scheduled task will ensure the visit queue entries are ended as well.

ojwanganto (approved):

corneliouzbett (approved):
Thanks! @mseaton 
I like the approach, it's more cleaner.

api/src/main/java/org/openmrs/module/queue/tasks/AutoCloseVisitQueueEntryTask.java
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * This iterates over all active VisitQueueEntries If the Visit associated with any of these has
+ * ended (non-null stopDatetime), then the QueueEntry associated with it is also ended at the same
+ * datetime as the Visit was stopped.
+ */
+@Slf4j
+public class AutoCloseVisitQueueEntryTask implements Runnable {
+	
+	private static boolean currentlyExecuting = false;
```
ibacher, line 31: Sorry for the post-commit review, but if this accessed concurrently (which seems to be the use case) it should be marked as `volatile`. Otherwise, if two threads that start close to one another, they might both see this value as `false` (or alternatively, but less importantly, as `true`).

PR #42, opened 2023-09-27
O3-2453 - remove where conditions from property JPA annotations

mseaton:


api/src/main/java/org/openmrs/module/queue/model/Queue.java
```diff
@@ -52,13 +55,59 @@ public class Queue extends BaseChangeableOpenmrsMetadata {
...
+	/**
+	 * @return all non-voided QueueEntries that do not start in the future and are not already ended
+	 */
+	public List<QueueEntry> getActiveQueueEntries() {
+		List<QueueEntry> l = new ArrayList<>();
+		Date now = new Date();
+		if (queueEntries != null) {
+			for (QueueEntry e : queueEntries) {
+				if (BooleanUtils.isNotTrue(e.getVoided())) {
+					if (!e.getStartedAt().after(now) && e.getEndedAt() == null) {
```
mseaton, line 72: Note:  the intention here was to exactly replicate the pre-existing logic.  I don't know if this is what we want or not (eg. we consider the possibility of "future" queue entries, and don't want entries that _start_ in the future to be considered active, but we don't consider the possibility of queue entries that have already started but whose endDate is in the future, but rather anything with an end date is inactive.  We can discuss the correctness of this logic, and happy to change this to whatever we come up with, but for the initial PR, just replicating existing behavior.

api/src/test/java/org/openmrs/module/queue/tasks/AutoCloseVisitQueueEntryTaskTest.java
```diff
@@ -7,11 +7,12 @@
  * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
```
mseaton, line 7: Note:  All the changes in this file are just due to the fact that I had put this in the wrong package, and I noticed this while creating a unit test for the Queue class, so this just moves the task unit test from the "model" to the "tasks" package.

omod/src/main/java/org/openmrs/module/queue/web/resources/QueueEntryCountSubResource.java
```diff
@@ -52,7 +52,7 @@ public void setParent(QueueEntryCount queueEntryCount, Queue queue) {
 	@Override
 	public PageableResult doGetAll(Queue queue, RequestContext requestContext) throws ResponseException {
 		return new GenericSingleObjectResult(Arrays.asList(new PropValue("queueName", queue.getName()),
-		    new PropValue("queueEntriesCount", queue.getQueueEntries().size())));
+		    new PropValue("queueEntriesCount", queue.getActiveQueueEntries().size())));
```
mseaton, line 55: This change is needed for this to replicate the existing functionality, but is this what we want?  Should the "getAll" resource limit itself to only active queue entries?  Or should getAll really ...get all.  And then we can limit this to only active entries by passing in a filter or search parameter.

omod/src/main/java/org/openmrs/module/queue/web/resources/QueueEntrySubResource.java
```diff
@@ -53,7 +53,7 @@ public void setParent(QueueEntry queueEntry, Queue queue) {
 	
 	@Override
 	public PageableResult doGetAll(Queue queue, RequestContext requestContext) throws ResponseException {
-		Collection<QueueEntry> queueEntries = queue.getQueueEntries();
+		Collection<QueueEntry> queueEntries = queue.getActiveQueueEntries();
```
mseaton, line 56: Same comment here as above.  This change is needed for this to replicate the existing functionality, but is this what we want?  Should the "getAll" resource limit itself to only active queue entries?  Or should getAll really ...get all.  And then we can limit this to only active entries by passing in a filter or search parameter.

omod/src/main/java/org/openmrs/module/queue/web/resources/QueueEntryCountSubResource.java
corneliouzbett: This is a weird resource. Just returns the counts. A response like:
```java
{
  "QueueName": "Triage Queue",
  "count": "36"
}
```
The idea is to get the number of patients in a queue. Mainly used in the UI to display totals for queues.

api/src/main/java/org/openmrs/module/queue/model/Queue.java
```diff
@@ -52,13 +55,59 @@ public class Queue extends BaseChangeableOpenmrsMetadata {
...
+						l.add(e);
+					}
+				}
+			}
+		}
+		return l;
```
corneliouzbett, line 78:
```suggestion
	public List<QueueEntry> getActiveQueueEntries() {
		List<QueueEntry> activeQueueEntries = new ArrayList<>();
		Date now = new Date();
		if (queueEntries != null) {
			for (QueueEntry queueEntry : queueEntries) {
				if (BooleanUtils.isNotTrue(queueEntry.getVoided())) {
					if (!queueEntry.getStartedAt().after(now) && queueEntry.getEndedAt() == null) {
						activeQueueEntries.add(queueEntry);
					}
				}
			}
		}
		return activeQueueEntries;
```

corneliouzbett (commented):
Some suggestions. Otherwise all good.

omod/src/main/java/org/openmrs/module/queue/web/resources/QueueEntrySubResource.java
corneliouzbett: Do we have a use case for that already? 
Yeah `getAll` should really get all queue entries and probably add more search parameters through search handlers.

omod/src/main/java/org/openmrs/module/queue/web/resources/QueueEntryCountSubResource.java
ibacher: Hmmm... It doesn't seem like this resource is going to perform in an ideal way. I'd prefer it if we had one endpoint that returned "the current queue", i.e., queue name, all active entries, and counts. Then we could use the custom view code if someone really want to limit things down to just requesting, e.g., the name and count. This just encourages a pattern of access where we're making a lot of REST requests, which is bad for performance.

@mseaton In the context of a queue, I think it's fine if the default is to return active entries only. It seems reasonable to have a query property that be used to also request the queue with inactive entries, but I'm not clear on the use-case for that. This may be a little different from other domains, but I tend to think of inactive queue entries as similar to retired metadata.

api/src/main/java/org/openmrs/module/queue/model/Queue.java
ibacher: Uh... shouldn't the logic here be something like:

```java
if (!queueEntry.getStartedAt().after(now) && (queueEntry.getEndedAt() == null || queueEntry.getEndedAt().before(now)))
```

omod/src/main/java/org/openmrs/module/queue/web/resources/QueueEntrySubResource.java
mseaton: I don't have an explicit use case yet.  It was more just a matter of wanting to ensure our API was flexible enough to handle these use cases if/when they come up and also that it makes intuitive sense and matches what users would expect.  Once we establish that this endpoint only returns active queue entries, if a use case _does_ come up where we want _all_ queue entries regardless of active or not, how would we handle that?  Would we have to pass in a parameter saying "includeInactive=true" for example?  Not the end of the world, just putting it out there for consideration as to how we'd want/expect our REST apis to work, and if there are existing conventions in OpenMRS that we honor those.

api/src/main/java/org/openmrs/module/queue/model/Queue.java
ibacher: Yeah, there's maybe some discussion to be had. I'm not exactly sure what the concrete use-case for future-dated entries is, but it seems like it's getting out of the realm of a "list of patients waiting for a resource".

omod/src/main/java/org/openmrs/module/queue/web/resources/QueueEntryCountSubResource.java
mseaton: I understand the primary use case.  A potentially more comprehensive endpoint might return an object that contains data that is a bit more complete and includes counts as the service/api deals with them, so it's clear.  For example, if our "getActiveQueueEntries" filters out voided, started-in-future, and ended entries (which it does), then we could ensure that all of these are represented in this endpoint comprehensively like:
```
{
  "QueueName": "Triage Queue",
  "totalCount": "63",
  "activeCount": "36",
  "futureCount": "5",
  "voidedCount": "2",
  "endedCount": "20"
}
```

ibacher (approved):

omod/src/main/java/org/openmrs/module/queue/web/resources/QueueEntryCountSubResource.java
mseaton: Sorry, I made this comment before I saw @ibacher  response.  Must have been typing at the same time.  You can disregard my comment, especially if we get rid of this endpoint as Ian seems to be suggesting :)

mseaton: > This may be a little different from other domains, but I tend to think of inactive queue entries as similar to retired metadata

I understand your point and I'm fine if the endpoint returns active entries by default.  I just wanted to bring it up to make sure.  As you indicate, we can always add support for query parameters like "includeEnded=true", "includeVoided=true", and "includeFuture=true" for those use cases if/when they arise.

mogoodrich: Yeah, haven't dug into it in much detail but the use cases for get all "ended" queue members or a count of all of them seems relatively small.  If I'm understand it correctly the "ended" queue entries will just continue to grow and grow indefinitely and the general need when asking for "all" would be "get me all currently in the queue".  There might be a case to say, "get me all cases when patient X was in this queue in the past month" but "getAllEntires" seems about as useful as "getAllObs".  (Again, If I'm understanding this all correctly).

PR #44, opened 2023-10-24
O3-2448 - Visit queue number customization

mseaton:
The goal of this PR is to address the issues raised in https://issues.openmrs.org/browse/O3-2448 and https://issues.openmrs.org/browse/O3-2465.  This PR does the following:

1. Adds patientQueueNumber / patient_queue_number to QueueEntry / queue_entry, in order to keep this information directly in the queue model and remove the need for a custom visit attribute.
2. Removes the bespoke code for generating queue numbers from within the core DAO and into an implementation behind an interface.  This enables this functionality to be off by default, added easily if needed, and customized however implementations want.
3. Removes the services and rest endpoints for generating and saving queue numbers as a standalone operation in favor of one that embeds this functionality in the visit queue number saving process.  This enables transactional behavior around generating and setting visit queue numbers, visit attributes, visit details, and queue entry details at the same time, rather than requiring consuming code orchestrate these steps together non-transactionally.

See specific comments throughout the PR.

I have this as a Draft PR initially to show where I've been heading and looking for feedback / suggestions.  This will obviously be disruptive and some work to get it working as expected in existing implementations that rely on current behavior.  It will also need to be paired with changes to the O3 frontend to remove the calls to generate visit numbers, etc.

@corneliouzbett / @CynthiaKamau / @ibacher interested in thoughts / feedback.  @mogoodrich FYI


api/src/main/java/org/openmrs/module/queue/processor/VisitAttributeQueueNumberGenerator.java
```diff
@@ -0,0 +1,87 @@
...
+				}
+			}
+		}
+		String paddedString = StringUtils.leftPad(String.valueOf(visitQueueNumber), 3, "0");
+		String serviceName = queue.getName().toUpperCase();
+		String prefix = serviceName.length() < 3 ? serviceName : serviceName.substring(0, 3);
+		String queueNumber = prefix + "-" + paddedString;
+		
+		// Associate this queue number directly
+		queueEntry.setPatientQueueNumber(queueNumber);
```
mseaton, line 70: This was not in the original implementation, but I see no harm in it, and I think for a generic front-end implementation we'll need this - the data for queue number in a known, fixed attribute of the queue entry - even if is also duplicated elsewhere (eg. a custom visit attribute).

api/src/main/java/org/openmrs/module/queue/processor/VisitAttributeQueueNumberGenerator.java
```diff
@@ -0,0 +1,87 @@
...
+		
+		// Create Visit Attribute using generated queue number
+		VisitAttribute visitAttribute = new VisitAttribute();
+		visitAttribute.setAttributeType(visitAttributeType);
+		visitAttribute.setValueReferenceInternal(queueNumber);
+		visitQueueEntry.getVisit().addAttribute(visitAttribute);
```
mseaton, line 76: It's possible / likely that with the new column, this is not longer behavior that is desired.  But I'll leave that to others to discuss and decide.

api/src/test/java/org/openmrs/module/queue/api/dao/QueueEntryDaoTest.java
```diff
@@ -244,16 +241,4 @@ public void shouldZeroCountQueueEntriesByBadStatus() {
 		assertThat(queueEntriesCountByStatusCount, notNullValue());
 		assertThat(queueEntriesCountByStatusCount, is(0L));
 	}
-
```
mseaton, line 247: This test was moved / reimplemented below.

api/src/test/resources/org/openmrs/module/queue/api/dao/QueueEntryDaoTest_conceptsInitialDataset.xml
```diff
@@ -13,8 +13,8 @@
     <concept concept_id="1000" retired="false" is_set="true" creator="1" date_created="2022-02-02 14:31:00.0" uuid="897eba27-2b38-43e8-91a9-4dfe3956a35t"/>
     <concept concept_id="1001" retired="false" is_set="false" creator="1" date_created="2022-02-02 14:40:00.0" uuid="90b910bd-298c-4ecf-a632-661ae2f446op"/>
     <concept concept_id="1002" retired="false" is_set="false" creator="1" date_created="2022-02-02 14:40:00.0" uuid="91b910bd-298c-4ecf-a632-661ae2f909ut"/>
```
mseaton, line 15: The below are bugs that I found - uuids and concept name ids that were re-used and not unique

mseaton:
Closing this PR for now as I'm rethinking this approach and working on other issues.

PR #45, opened 2023-10-25
Remove VisitQueueEntry as a separate entity in favor of an optional p…

mseaton:
…roperty on QueueEntry; add additional searching capabilities; adjust web service resources to reflect changes.


mseaton:
This is admittedly going to be very hard to review, there are a lot of changes, and it is likely to grow as I add more commits on top of this.  The main things this PR aims to achieve:

* There was a lot of logic duplicated between VisitQueueEntry and QueueEntry, and in some cases inconsistencies and/or capabilities in one but not the other.  The data model was also a bit non-standard.  To simplify things, the entire `VisitQueueEntry` domain and all of the associated services, daos, resources, validators, tests, etc. have been removed in favor of adding an optional `visit` property to QueueEntry, and then expanding the capabilities of the services, daos, resources, validators, and tests to support working with QueueEntry objects that are and are not associated with a visit.

*There were some issues previously identified with the QueueEntry and VisitQueueEntry services, daos, and resources that this provided an opportunity to address as functionality was consolidated across these domains.  Namely, issues identified in O3-2441 and O3-2447.  This addesses many of these and also expands on the searching capabilities available in the QueueEntryService/DAO to enable searching across a wide range of QueueEntry properties (including whether a QueueEntry is associated with a visit).

* There was previously a top-level `visit-queue-entry` web service resource, but for `queue-entry` this was only available as a subresource.  I have repurposed the `visit-queue-entry` resource into a general purpose `queue-entry` resource to enable querying for these independently from the queue.  In this process, there were a number of issues I identified in the existing web services, which led me to remove several of them.  This should be reviewed:
  ** InActiveQueueEntrySearchHandler - redundant, as I repurposed in favor of the new QueueEntry resource which can be used to retrieve queue entries based on active/inactive, either associated with a specific queue or not.
  ** QueueCountSearchHandler (SubResourceSearchHandler of queue).  This does not seem like it was implemented correctly, as it previously returned all queue entries by status, not just those limited to a specific queue.  This is also redundant to the QueueEntryMetricsResource with appropriate search parameters.
  ** QueueEntrySearchHandler (SubResourceSearchHandler of queue).  This does not seem like it was implemented correctly, as it previously returned all queue entries by status, not just those limited to a specific queue.  This is also redundant to the QueueEntryResource with appropriate search parameters.

Still TODO:
* Broaden unit test coverage for queue entry search method
* Add broader search capabilities to the queue entry web service resource and test these
* Determine whether the `QueueEntryMetricsResource` is best practice or if we can/should just use the `QueueEntryResource` and return the metrics when it is a HEAD request.  Interested in thoughts @ibacher .  If we keep it, we should improve the way the metrics are presented/returned and expand the search capabilities to mimic the `QueueEntryResource`

ibacher:
> Determine whether the QueueEntryMetricsResource is best practice or if we can/should just use the QueueEntryResource and return the metrics when it is a HEAD request. Interested in thoughts @ibacher . If we keep it, we should improve the way the metrics are presented/returned and expand the search capabilities to mimic the QueueEntryResource

`QueueEntryMetricsResource` sounds misnamed. We might have metrics-per-queue (e.g., number of active members in queue, number of all-time members in queue, average active wait time, average overall wait time, max wait time, etc.), but not per-queue entry (which doesn't seem to be what that's doing anyway.

I'm not sure I follow the idea of replacing that with a HEAD call... HTTP HEAD is supposed to return the same as HTTP GET except without the body, so just the headers and I don't think we should break that contract. That said, making metrics into a subresource of a Queue so one could hit `/queue/<UUID>/metrics` would seem to make sense. Pure counts of currently waiting queue members should probably just be a standard property of the `/queue` resource... The idea with "metrics" was to track more longitudinal / process-oriented things (how long do patients wait for X)? If we're not yet using that or don't really have a use-case for it, I'd leave it out altogether.

While I haven't read through the PR in detail, every bullet point above sounds exactly right and I generally trust that you've implemented things sanely.

mseaton:
> I'm not sure I follow the idea of replacing that with a HEAD call... HTTP HEAD is supposed to return the same as HTTP GET except without the body, so just the headers and I don't think we should break that contract.

Thanks @ibacher .  I thought I remember you commenting about this in the recent past, either here or in a different domain, but I can't pinpoint it.  In any event, the issue is that there are basically 2 parallel things that are being supported:

* Thing 1:  Service method + DAO method + REST resource to get queue entries that match XYZ criteria
* Thing 2:  Service method + DAO method + REST resource to get the _count_ of queue entries that match XYZ criteria

The services and DAOs exist for Thing 2 because they are presumably much faster to compute than Thing 1 (eg. `select count(*) from foo` is faster than `select fully_hydrated_object from foo;` + return these as a list + return the count of this list.  I don't know if this speed difference is meaningful or not, but it seems like it could be

The question I posed here is whether it is best practice to have a separate REST resource for this, whether it is best practice to use the same REST resource with a different method (eg. HEAD), or whether it is best practice to use the same REST resource with a different parameter/flag.  This [question on stack overflow](https://stackoverflow.com/questions/47906473/rest-api-should-we-have-separate-api-for-result-and-result-count) (and what I thought I read in one your your PR reviews somewhere @ibacher ) is where the HEAD concept came from.

Basically - if a HEAD request is intended to return a count of results, but not the actual results, this seems like basically what we are looking for, especially if - behind the scenes in the REST controllers - we can service this with the more efficient service/dao method that retrieves only the count of results rather than returning a collection of all of the full result objects.

api/src/main/java/org/openmrs/module/queue/api/dao/impl/QueueEntryDaoImpl.java
```diff
@@ -41,84 +27,49 @@ public QueueEntryDaoImpl(@Qualifier("sessionFactory") SessionFactory sessionFact
...
-		criteria.add(Property.forName("qe.status")
-		        .in(conceptByNameDetachedCriteria(conceptStatus, localePreferred, conceptNameType)));
+	public Long getCountOfQueueEntries(QueueEntrySearchCriteria searchCriteria) {
+		Criteria criteria = createCriteriaFromSearchCriteria(searchCriteria);
 		criteria.setProjection(Projections.rowCount());
-		
 		return (Long) criteria.uniqueResult();
 	}
 	
-	@Override
```
CynthiaKamau, line 77: I think we should retain this before we decide the way forward

api/src/main/java/org/openmrs/module/queue/api/dao/impl/QueueEntryDaoImpl.java
```diff
@@ -41,84 +27,49 @@ public QueueEntryDaoImpl(@Qualifier("sessionFactory") SessionFactory sessionFact
...
-	public String generateVisitQueueNumber(Location location, Queue queue) {
```
CynthiaKamau, line 78: I think we should retain this before we decide the way forward

mseaton: Thanks @CynthiaKamau - this functionality wasn't removed in this PR, it was just moved into entirely into the service.  So the existing visit queue number functionality remains as it was, just not using a custom DAO method to do it.

api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
```diff
@@ -115,20 +127,34 @@ public String generateVisitQueueNumber(Location location, Queue queue, Visit vis
...
+		criteria.setLocations(Collections.singletonList(location));
+		Date onOrAfter = Date.from(LocalDateTime.now().with(LocalTime.MIN).atZone(ZoneId.systemDefault()).toInstant());
+		criteria.setStartedOnOrAfter(onOrAfter);
+		Date onOrBefore = Date.from(LocalDateTime.now().with(LocalTime.MAX).atZone(ZoneId.systemDefault()).toInstant());
+		criteria.setStartedOnOrAfter(onOrBefore);
+		Long nextQueueNumber = getCountOfQueueEntries(criteria) + 1;
+		String paddedString = StringUtils.leftPad(String.valueOf(nextQueueNumber), 3, "0");
+		String serviceName = queue.getName().toUpperCase();
+		String prefix = serviceName.length() < 3 ? serviceName : serviceName.substring(0, 3);
+		String queueNumber = prefix + "-" + paddedString;
```
mseaton, line 142: @CynthiaKamau - see this section above.  This basically replaces what was previously being done in a custom DAO method to instead now be using the generic search DAO function to achieve the same result.

api/src/main/resources/liquibase.xml
```diff
@@ -385,4 +385,38 @@
...
+    </changeSet>
+
+    <!-- TODO: Uncomment this to remove the legacy visit_queue_entries table
+    <changeSet id="drop_visit_queue_entries_20231025" author="mseaton">
+        <preConditions onFail="MARK_RAN">
+            <tableExists tableName="visit_queue_entries"/>
+        </preConditions>
+        <dropTable tableName="visit_queue_entries"/>
+    </changeSet>
+    -->
```
mseaton, line 421: Note that ideally we'd drop the visit_queue_entries table entirely to eliminate any confusion and keep things cleaned up, but I have this commented out for now just to be on the safe side as we consider this PR overall.

PR #47, opened 2023-11-22
Add back createQueue method for backwards-compatibility

mseaton:


mseaton:
@dkayiwa - I see you released 2.0.0 recently, but @mogoodrich just discovered this issue whereby the Initializer module is using a method I had removed in version 2.0.0.  In order to better support this, this PR adds this method back in.  We likely want to do a 2.0.1 maintenance release and update distros to use that instead.

mseaton:
No idea why the github action build is failing, but it didn't start with this commit, and it builds cleanly without errors for me locally, so it should be safe to ignore I think (though we should investigate)

mogoodrich (approved):
LGTM!

PR #48, opened 2023-12-01
UHM-7557, prevent duplicate queue entries

cioan:
@mseaton , just wanted to see if this is the right direction to fix the problem with the duplicate queue entries? I could add unit tests too, but first wanted to make sure we want to fix this at the service layer or we could also add some checks in the rest controller. Thanks!

### DIFF
--- a/api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
+++ b/api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
@@ -83,6 +83,12 @@ public class QueueEntryServiceImpl extends BaseOpenmrsService implements QueueEn
 				throw new IllegalArgumentException("Patient mismatch - visit.patient does not match patient");
 			}
 		}
+		QueueEntrySearchCriteria searchCriteria = new QueueEntrySearchCriteria();
+		searchCriteria.setPatient(queueEntry.getPatient());
+		searchCriteria.setQueues(Collections.singletonList(queueEntry.getQueue()));
+		if (getQueueEntries(searchCriteria).size() > 0) {
+			throw new APIException("Patient already in the queue");
+		}
 		return dao.createOrUpdate(queueEntry);
 	}
 	


api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
```diff
@@ -83,6 +83,12 @@ public QueueEntry saveQueueEntry(QueueEntry queueEntry) {
 				throw new IllegalArgumentException("Patient mismatch - visit.patient does not match patient");
 			}
 		}
+		QueueEntrySearchCriteria searchCriteria = new QueueEntrySearchCriteria();
+		searchCriteria.setPatient(queueEntry.getPatient());
+		searchCriteria.setQueues(Collections.singletonList(queueEntry.getQueue()));
+		if (getQueueEntries(searchCriteria).size() > 0) {
+			throw new APIException("Patient already in the queue");
+		}
```
mseaton, line 91: This is the right direction @cioan .  A few things that are needed:

1. Queue Entries are never deleted, they just end.  So patients will be expected to be on the same Queue multiple times.  That could be on the same day, or it could be some arbitrary point in the future.  What needs to be prevented is for them to be on the same Queue at the same time actively (based on started_at and ended_at properties).

2. One the UI-side we likely want to display a nice error message to the user explaining why they couldn't add a patient to a queue.  This particular exception is both too generic (APIException) and too non-localized, to be able to drive the UI.  We need to improve this such that it can communicate back to the UI, via the REST endpoints, what the specific error is, and what the localized display text of this error message should be.

mseaton (changes requested):

api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
cioan: Thanks @mseaton for the key points above and for the explanation yesterday on how the queue entries could overlap. I just commited a few more changes on my second attempt to prevent duplicate queue entries within the same timeframe. Hope this cover all possible use cases. Thanks!

api/src/main/java/org/openmrs/module/queue/api/DuplicateQueueEntryException.java
```diff
@@ -0,0 +1,43 @@
...
+	private static final long serialVersionUID = 1L;
+	
+	public DuplicateQueueEntryException() {
+	}
+	
+	/**
+	 * @param message
+	 */
+	public DuplicateQueueEntryException(String message) {
+		super(Context.getMessageSourceService().getMessage(message));
```
mseaton, line 26: Unfortunately, we should not put calls to the Context like this within this constructor.  We need to pass in a pre-translated message, not translate within here.

api/src/main/resources/messages_es.properties
```diff
@@ -0,0 +1,11 @@
...
+# This Source Code Form is subject to the terms of the Mozilla Public License,
+# v. 2.0. If a copy of the MPL was not distributed with this file, You can
+# obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+# the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+#
+# Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+# graphic logo is a trademark of OpenMRS Inc.
+#
+
+queue.entry.duplicate.patient=Patient already in the queue
```
mseaton, line 11: Either provide reasonable default translations in the appropriate language or remove the non-english translation files from the module.  They can always be translated downstream via Initializer.

api/src/test/java/org/openmrs/module/queue/api/QueueEntryServiceTest.java
```diff
@@ -109,6 +114,62 @@ public void shouldCreateNewQueueEntryRecord() {
...
+		when(secondQueueEntry.getStatus()).thenReturn(conceptStatus);
+		when(secondQueueEntry.getPriority()).thenReturn(conceptPriority);
+		when(secondQueueEntry.getPatient()).thenReturn(patient);
+		when(secondQueueEntry.getStartedAt()).thenReturn(secondQueueStartDate);
+
+		QueueEntrySearchCriteria searchCriteria = new QueueEntrySearchCriteria();
+		searchCriteria.setPatient(secondQueueEntry.getPatient());
+		searchCriteria.setQueues(Collections.singletonList(queue));
+		//tell the dao to return the first queueEntry that was successfully saved above
+		when(dao.getQueueEntries(searchCriteria)).thenReturn(Collections.singletonList(result));
```
mseaton, line 159: This doesn't actually test the most meaningful part of what needs testing - namely that the date overlapping logic works as intended.  Hopefully moving that out to a separate method will facilitate that.

api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
mseaton: Please add unit tests that demonstrate that all possible use cases are in fact covered.  This should include a new queue entry that is fully before, fully within, and fully after an existing queue entry, for both queue entries that are ended and non-ended.  It should also cover edge cases (eg. when dates on either bound are equal to each other).

api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
```diff
@@ -83,6 +84,19 @@ public QueueEntry saveQueueEntry(QueueEntry queueEntry) {
...
+		searchCriteria.setPatient(queueEntry.getPatient());
+		searchCriteria.setQueues(Collections.singletonList(queueEntry.getQueue()));
+		List<QueueEntry> queueEntries = getQueueEntries(searchCriteria);
+		for (QueueEntry entry : queueEntries) {
+			if (!((queueEntry.getStartedAt().before(entry.getStartedAt())
+			        && (queueEntry.getEndedAt() != null && queueEntry.getEndedAt().before(entry.getStartedAt())))
+			        || ((entry.getEndedAt() != null) && (queueEntry.getStartedAt().after(entry.getEndedAt()))))) {
+				//if the new queueEntry endDate is NOT before the entry startDate OR the new queueEntry startDate is NOT after the entry endDate,
+				//then the new queueEntry overlaps with the existing entry
+				throw new DuplicateQueueEntryException("queue.entry.duplicate.patient");
```
mseaton, line 97: Will need to add the messageSourceService to this, wire it up via the application context xml, and then use it here to do the necessary translation.

api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
```diff
@@ -83,6 +84,19 @@ public QueueEntry saveQueueEntry(QueueEntry queueEntry) {
 				throw new IllegalArgumentException("Patient mismatch - visit.patient does not match patient");
 			}
 		}
+		QueueEntrySearchCriteria searchCriteria = new QueueEntrySearchCriteria();
+		searchCriteria.setPatient(queueEntry.getPatient());
+		searchCriteria.setQueues(Collections.singletonList(queueEntry.getQueue()));
+		List<QueueEntry> queueEntries = getQueueEntries(searchCriteria);
+		for (QueueEntry entry : queueEntries) {
```
mseaton, line 91: I feel like we would be best served if the logic below was extracted out.  This could potentially be a method directly on QueueEntry (eg. `boolean isOverlapping(QueueEntry)`) or could be something we add to QueueUtils (eg. `static boolean datesOverlap(Date startDate1, Date endDate1, Date startDate2, Date endDate2)`.  Then, we can add a lot of unit tests around this method to ensure that it meets all of the right behavior, and then when we test the actual service, we can potentially need to cover fewer of the various use cases.

mseaton (changes requested):

api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
```diff
@@ -83,6 +85,17 @@ public QueueEntry saveQueueEntry(QueueEntry queueEntry) {
 				throw new IllegalArgumentException("Patient mismatch - visit.patient does not match patient");
 			}
 		}
+		QueueEntrySearchCriteria searchCriteria = new QueueEntrySearchCriteria();
+		searchCriteria.setPatient(queueEntry.getPatient());
+		searchCriteria.setQueues(Collections.singletonList(queueEntry.getQueue()));
+		List<QueueEntry> queueEntries = getQueueEntries(searchCriteria);
+		for (QueueEntry qe : queueEntries) {
```
mseaton, line 92: @cioan - note that this was missing and I've added it in - that we need to allow re-saving a saved queue entry.  Without this you'd get a validation error trying to save / update the same queue entry.

api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
```diff
@@ -83,6 +85,17 @@ public QueueEntry saveQueueEntry(QueueEntry queueEntry) {
...
+			if (!qe.equals(queueEntry)) {
+				if (QueueUtils.datesOverlap(qe.getStartedAt(), qe.getEndedAt(), qe.getStartedAt(), qe.getEndedAt())) {
+					throw new DuplicateQueueEntryException("queue.entry.duplicate.patient");
```
mseaton, line 95: @cioan - note, I removed the translation of the exception altogether.  This can be handled downstream in the view layer.

api/src/main/java/org/openmrs/module/queue/exception/DuplicateQueueEntryException.java
```diff
@@ -0,0 +1,42 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module.queue.exception;
```
mseaton, line 10: @cioan - note, I moved this to a new package

api/src/main/java/org/openmrs/module/queue/utils/QueueUtils.java
```diff
@@ -74,4 +74,28 @@ public static double computeAverageWaitTimeInMinutes(List<QueueEntry> queueEntri
 		}
 		return averageWaitTime;
 	}
+	
+	/**
+	 * @param startDate1, endDate1 - the start and end date of one timeframe
+	 * @param startDate2, endDate2 - the start and end date of second timeframe
+	 * @return boolean - indicating whether the timeframes overlap
+	 */
+	public static boolean datesOverlap(Date startDate1, Date endDate1, Date startDate2, Date endDate2) {
```
mseaton, line 83: @cioan - I rewrote this implementation, as there were several use cases that it didn't seem to cover.  Also, I updated the handling of nulls so that they indicate an unbounded condition (eg. overlapping nulls should be overlapping, not non-overlapping).  Have a look at this new implementation and let me know if you see any issues.

api/src/test/java/org/openmrs/module/queue/api/QueueEntryServiceTest.java
```diff
@@ -109,6 +109,48 @@ public void shouldCreateNewQueueEntryRecord() {
 		assertThat(result.getPriority(), is(conceptPriority));
 	}
 	
+	@Test
+	public void shouldNotCreateDuplicateOverlappingQueueEntryRecords() {
+		Queue queue = new Queue();
```
mseaton, line 114: @cioan - note, I removed all of the mocking in favor of just creating and populating the actual objects, they are just beans.

api/src/test/java/org/openmrs/module/queue/api/QueueEntryServiceTest.java
```diff
@@ -109,6 +109,48 @@ public void shouldCreateNewQueueEntryRecord() {
...
+		QueueEntrySearchCriteria searchCriteria = new QueueEntrySearchCriteria();
+		searchCriteria.setPatient(patient);
+		searchCriteria.setQueues(Collections.singletonList(queue));
+		
+		when(dao.createOrUpdate(savedQueueEntry)).thenReturn(savedQueueEntry);
+		when(dao.getQueueEntries(searchCriteria)).thenReturn(Collections.singletonList(savedQueueEntry));
+		
+		// Should be able to save and re-save a queue entry without causing validation failure
+		savedQueueEntry = queueEntryService.saveQueueEntry(savedQueueEntry);
+		queueEntryService.saveQueueEntry(savedQueueEntry);
```
mseaton, line 142: @cioan - note, I added this test to ensure that saving the same queue entry multiple times doesn't fail validation

api/src/test/java/org/openmrs/module/queue/utils/QueueUtilsTest.java
```diff
@@ -0,0 +1,55 @@
...
+	
+	private static final Date AUG_2 = QueueUtils.parseDate("2023-08-02 10:00:00");
+	
+	private static final Date AUG_3 = QueueUtils.parseDate("2023-08-03 10:00:00");
+	
+	private static final Date AUG_4 = QueueUtils.parseDate("2023-08-04 10:00:00");
+	
+	@Test
+	public void shouldReturnTrueIfDatesOverlap() {
+		// Test that nulls are handled as open-ended dates
```
mseaton, line 33: @cioan - see how I rewrote this test.  I found this style a lot easier for me to follow, and more conducive to adding lots of test criteria.  Have a look and let me know if you think any test cases are missing, and if you have questions about whether certain conditions are handled correctly.

cioan:
Thanks @mseaton ! LGTM. Can I merge in this PR? Thanks!

PR #49, opened 2024-01-02
O3-2446 - Queue module service, status, and priority configuration

mseaton:
See:  https://openmrs.atlassian.net/browse/O3-2446
* Updates the global properties that configure the available/default services, priorities, and statuses for queues to allow configuration by uuid or mapping:code in addition to the ( currently supported ) concept name.
* Adds new priorityConceptSet and serviceConceptSet to the Queue object and data model
* Adds new services to retrieve the allowed priorities and services for a given queue, first using any specific concept set configured on the queue, and defaulting to what is configured in the (existing) global property
* Updates the validators to respect this new configuration
* Updates the REST endpoint for Queue to support these new properties and to add a convenience property to retrieve the allowed priorities and statuses for a given Queue.


api/src/main/java/org/openmrs/module/queue/api/QueueServicesWrapper.java
```diff
@@ -104,14 +110,13 @@ public Concept getConcept(String conceptRef) {
...
-		if (concepts.size() == 1) {
-			return concepts.get(0);
-		} else if (concepts.size() > 1) {
-			throw new IllegalArgumentException("More than one concept is found with name: " + conceptRef);
+		//handle name.  this isn't ideal, as core will just log a warning if there are 2 concepts with the same name
+		//but there are no alternative suitable methods to get a List of Concepts that exactly match a given name
+		c = getConceptService().getConceptByName(conceptRef);
+		if (c == null) {
+			throw new IllegalArgumentException("Unable to find concept: " + conceptRef);
 		}
```
mseaton, line 118: I discovered in testing that the method I was previously using (`getConceptService().getConceptsByName(conceptRef)`)  does a fuzzy-match (eg. partial match) rather than an exact match, so it was matching on too many concepts.  Unfortunately, there is no public method in the ConceptService to get back all Concepts that exactly match a particular name, and the one above that I am using will just log a warning if multiple names are found and return the first one, which really isn't ideal, but given that it is also what is used in the new Core 2.6 method that accomplishes the same as the above, I went with it.  We likely should address this in core, and provide some additional (better) options.

api/src/main/java/org/openmrs/module/queue/validators/QueueValidator.java
```diff
@@ -38,38 +35,22 @@ public void validate(Object target, Errors errors) {
...
-	 * @param location location to check if it exists
-	 * @return true or false if the location exists
-	 */
-	public boolean isValidLocation(@NotNull Location location) {
-		if (location == null) {
-			return false;
-		}
-		Location isExistentLocation = Context.getLocationService().getLocationByUuid(location.getUuid());
-		return isExistentLocation != null;
-	}
```
mseaton, line 73: This seemed rather pointless.  We already check for null earlier in the method, and we don't typically check whether a nested property is actually saved or not, so I removed this check.

Per my added comment above, we _do_ probably want to validate that the location is actually tagged as a Queue Location, but I'll leave that for a separate ticket if desired.

omod/src/main/java/org/openmrs/module/queue/web/resources/QueueResource.java
```diff
@@ -156,6 +166,16 @@ public String getDisplay(Queue queue) {
...
 	
+	@PropertyGetter("allowedPriorities")
+	public Object getAllowedPriorities(Queue delegate) {
+		return services.getAllowedPriorities(delegate);
+	}
+	
+	@PropertyGetter("allowedStatuses")
+	public Object getAllowedStatuses(Queue delegate) {
+		return services.getAllowedStatuses(delegate);
+	}
```
mseaton, line 177: I'm not entirely sure if this is the right way to support this, but adding these to the Queue resource seems the most straightforward and helpful.

api/src/main/java/org/openmrs/module/queue/api/QueueServicesWrapper.java
ibacher: IIRC, the method in core was basically pulled straight from HTMLFormEntry and similarly didn't want to mess with what already worked. I agree having something like `getConceptsByExactName()` in core would actually be really, really useful for these cases where we're using concept names as concept identifiers.

api/src/main/java/org/openmrs/module/queue/api/QueueServicesWrapper.java
```diff
@@ -257,4 +262,54 @@ public QueueRoom getQueueRoom(String queueRoomRef) {
...
 		throw new IllegalArgumentException("Unable to find queueRoom: " + queueRoomRef);
 	}
+	
+	/**
+	 * @return the allowed service concepts
+	 */
+	public List<Concept> getAllowedServices() {
```
ibacher, line 269: Is there a reason that this isn't also a property of queues?

ibacher (approved):
LGTM! Thanks @mseaton!

lluismf:
It's kind of weird to throw IllegalArgumentException if the concept does not exist.
I'd rather throw ObjectNotFoundException or return null.

PR #52, opened 2024-02-13
O3-2862 - Queue Module - validate that queue entry falls within visit…

mseaton:
… when saving visit


ibacher (approved):

ibacher (approved):

api/src/main/java/org/openmrs/module/queue/validators/VisitWithQueueEntriesValidator.java
```diff
@@ -0,0 +1,77 @@
...
+		}
+		Visit visit = (Visit) target;
+		
+		// This implementation is copied to match the core validator approach to nested encounters
+		if (visit.getId() != null) {
+			Date startDateTime = visit.getStartDatetime();
+			Date stopDateTime = visit.getStopDatetime();
+			
+			QueueEntrySearchCriteria criteria = new QueueEntrySearchCriteria();
+			criteria.setIsEnded(null);
```
mogoodrich, line 53: I assume setting this to null means returns all queues, both those that are ended and those that aren't ended.

mogoodrich (approved):
LGTM, thanks @mseaton ... nice that you can add in validators for existing objects this way.

api/src/main/java/org/openmrs/module/queue/validators/VisitWithQueueEntriesValidator.java
mseaton: Correct, the default (which I don't love, but the way it is implemented) is to only return active entries by default.  So to get all entries, you need to explicitly null out this property :/

mogoodrich (approved):
LGTM, thanks @mseaton

PR #54, opened 2024-02-15
O3-2868: Provide service method and REST endpoint to facilitate transitioning from one queue entry to another

mseaton:


omod/src/main/java/org/openmrs/module/queue/web/QueueEntryTransitionRestController.java
```diff
@@ -0,0 +1,110 @@
...
+	
+	private final QueueServicesWrapper services;
+	
+	@Autowired
+	public QueueEntryTransitionRestController(QueueServicesWrapper services) {
+		this.services = services;
+	}
+	
+	@RequestMapping(method = { RequestMethod.PUT, RequestMethod.POST })
+	@ResponseStatus(HttpStatus.OK)
```
mseaton, line 58: I copied the skeleton of this implementation from the `PasswordResetController2_2` class in the webservices.rest module, looking for a pattern to follow.  Interested if we think this should return a response body or not (eg. containing the newly created queue entry, etc)

api/src/main/java/org/openmrs/module/queue/api/QueueEntryService.java
```diff
@@ -48,6 +49,15 @@ public interface QueueEntryService {
...
 	QueueEntry saveQueueEntry(@NotNull QueueEntry queueEntry);
 	
+	/**
+	 * Transitions a queue entry by ending one queue entry and creating a new queue entry that starts at
+	 * that time
+	 * 
+	 * @param queueEntryTransition the queueEntryTransition
+	 * @return the new QueueEntry that is created
+	 */
+	QueueEntry transitionQueueEntry(@NotNull QueueEntryTransition queueEntryTransition);
```
mseaton, line 59: Is this the right return type?  Should `QueueEntryTransition` have an additional property for the new queueEntry (which starts out null), and then this method would return the modified `QueueEntryTransition`?

api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
```diff
@@ -100,6 +101,20 @@ public QueueEntry saveQueueEntry(QueueEntry queueEntry) {
...
 	
+	/**
+	 * @see QueueEntryService#transitionQueueEntry(QueueEntryTransition)
+	 */
+	@Override
+	public QueueEntry transitionQueueEntry(QueueEntryTransition queueEntryTransition) {
+		// End the initial queue entry
+		QueueEntry queueEntryToStop = queueEntryTransition.getQueueEntryToTransition();
+		queueEntryToStop.setEndedAt(queueEntryTransition.getTransitionDate());
+		saveQueueEntry(queueEntryToStop);
```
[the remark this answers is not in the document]
ibacher: Unfortunately, the limitation this will bypass the SaveHandlers and any other AOP. The reason is that AOP is implemented (in Spring) using an AOP proxy, but since this results in a direct method call, the proxy can't intercept the call (see [the Spring docs](https://docs.spring.io/spring-framework/docs/5.3.32/reference/html/core.html#aop-understanding-aop-proxies)).

Instead of the call to `Context.getService()` it might be slightly more testable to use Spring's `AopContext`:

```java
((QueueEntryService) AopContext.currentProxy()).saveQueueEntry(queueEntryToStop);
```

This has the advantage that, in test code, you can just write:

```java
AopContext.setCurrentProxy(queueEntryService);
```

omod/src/main/java/org/openmrs/module/queue/web/QueueEntryTransitionRestController.java
```diff
@@ -0,0 +1,110 @@
...
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.ResponseStatus;
+
+/**
+ * The main controller that exposes additional end points for order entry
+ */
+@Controller
+@RequestMapping(value = "/rest/" + RestConstants.VERSION_1 + "/queue-entry-transition")
```
ibacher, line 37: I think we generally try to namespace the REST stuff by moduleId:

```suggestion
@RequestMapping(value = "/rest/" + RestConstants.VERSION_1 + "/queue/queue-entry-transition")
```

omod/src/main/java/org/openmrs/module/queue/web/QueueEntryTransitionRestController.java
ibacher: I think the response here (and in the PasswordResetController) should be `HttpStatus.NO_CONTENT` if there's no body.

That said, for things like this where we're effectively creating a whole new entry, I do think it's better if the response returns that entry... That gives us a lot more flexibility on the frontend and allows us to potentially avoid an unnecessary GET request.

omod/src/main/java/org/openmrs/module/queue/web/QueueEntryTransitionRestController.java
```diff
@@ -0,0 +1,110 @@
...
+	public void transitionQueueEntry(@RequestBody Map<String, String> body) {
+		QueueEntryTransition transition = new QueueEntryTransition();
+		
+		// Queue Entry to Transition
+		String queueEntryUuid = body.get(QUEUE_ENTRY_TO_TRANSITION);
+		Optional<QueueEntry> queueEntryOptional = services.getQueueEntryService().getQueueEntryByUuid(queueEntryUuid);
+		if (!queueEntryOptional.isPresent()) {
+			throw new APIException(QUEUE_ENTRY_TO_TRANSITION + " is a required parameter");
+		}
+		transition.setQueueEntryToTransition(queueEntryOptional.get());
```
ibacher, line 68: I think the preferred unwrapping pattern is something like:

```java
QueueEntry queueEntry = queueEntryOptional.getOrThrow(() -> new APIException());
```

api/src/main/java/org/openmrs/module/queue/api/QueueEntryService.java
ibacher: I think this is a lot cleaner that returning the `QueueEntryTransition`. It's also a good way of demarcating that the transition is over at this point.

ibacher (changes requested):
Thanks @mseaton! This is really good work, but a few things that I think need to be adjusted.

api/src/main/java/org/openmrs/module/queue/api/QueueEntryService.java
ojwanganto: +1

api/src/main/java/org/openmrs/module/queue/model/QueueEntryTransition.java
```diff
@@ -0,0 +1,55 @@
...
+
+import lombok.Data;
+import org.openmrs.Concept;
+
+/**
+ * Bean definition that encapsulates the supported criteria for saving a direct transition from one
+ * QueueEntry to another QueueEntry.
+ */
+@Data
+public class QueueEntryTransition implements Serializable {
```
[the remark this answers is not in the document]
ojwanganto: The approach is a good one

omod/src/main/java/org/openmrs/module/queue/web/QueueEntryTransitionRestController.java
mogoodrich: Yeah, my gut would be to return the new entry.

api/src/main/java/org/openmrs/module/queue/model/QueueEntryTransition.java
```diff
@@ -0,0 +1,55 @@
...
+		queueEntry.setQueue(newQueue == null ? queueEntryToTransition.getQueue() : newQueue);
+		queueEntry.setPatient(queueEntryToTransition.getPatient());
+		queueEntry.setVisit(queueEntryToTransition.getVisit());
+		queueEntry.setPriority(newPriority == null ? queueEntryToTransition.getPriority() : newPriority);
+		queueEntry.setPriorityComment(queueEntryToTransition.getPriorityComment());
+		queueEntry.setStatus(newStatus == null ? queueEntryToTransition.getStatus() : newStatus);
+		queueEntry.setSortWeight(queueEntryToTransition.getSortWeight());
+		queueEntry.setLocationWaitingFor(queueEntryToTransition.getLocationWaitingFor());
+		queueEntry.setProviderWaitingFor(queueEntryToTransition.getProviderWaitingFor());
+		queueEntry.setQueueComingFrom(queueEntryToTransition.getQueue());
```
mogoodrich, line 51: The practice I assume is that this is supposed to be populated even if it's coming from the new queue?

api/src/main/java/org/openmrs/module/queue/model/QueueEntryTransition.java
```diff
@@ -0,0 +1,55 @@
...
+		QueueEntry queueEntry = new QueueEntry();
+		queueEntry.setQueue(newQueue == null ? queueEntryToTransition.getQueue() : newQueue);
+		queueEntry.setPatient(queueEntryToTransition.getPatient());
+		queueEntry.setVisit(queueEntryToTransition.getVisit());
+		queueEntry.setPriority(newPriority == null ? queueEntryToTransition.getPriority() : newPriority);
+		queueEntry.setPriorityComment(queueEntryToTransition.getPriorityComment());
+		queueEntry.setStatus(newStatus == null ? queueEntryToTransition.getStatus() : newStatus);
+		queueEntry.setSortWeight(queueEntryToTransition.getSortWeight());
+		queueEntry.setLocationWaitingFor(queueEntryToTransition.getLocationWaitingFor());
+		queueEntry.setProviderWaitingFor(queueEntryToTransition.getProviderWaitingFor());
```
mogoodrich, line 50: Seems kind of strange that the location and provider wouldn't get reset if the queue is changing?  (But probably just my misunderstanding of how queues are supposed to be used)

api/src/main/java/org/openmrs/module/queue/api/QueueEntryService.java
mogoodrich: +1

api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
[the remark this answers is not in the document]
mogoodrich: Ugh, yeah, this is a pain... :)

mogoodrich (approved):
Generally LGTM, though we should make sure we switch to go through the Context for AOP, etc...

I added a couple comments/questions on QueueEntryTransition, but those are really more for my own understandng of how queues work.  Thanks @mseaton !

api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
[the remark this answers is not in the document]
mseaton: OK.  I'll just make my unit test into a context sensitive one, though that's a cool trick @ibacher .  I had written it the way I was supposed to originally, but then changed when my unit tests failed.  Just hoping maybe this had been magically fixed I guess.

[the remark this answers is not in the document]
mseaton: Actually I found a different approach that seemed to work out easier.  See last commit

omod/src/main/java/org/openmrs/module/queue/web/QueueEntryTransitionRestController.java
mseaton: Done.  I went with the REF representation, which seems sufficient.

omod/src/main/java/org/openmrs/module/queue/web/QueueEntryTransitionRestController.java
mseaton: Yes, 100% @ibacher , but in this case we already have a resource at `/queue`, it is what returns all of the queues, and specific resources at `/queue/<uuid>` which returns a specific queue.  So that would seem to prevent us from moving to a `queue` namespace unfortunately.  For now, what I've done is consistent with the rest of the module that exists, and I'm trying to maintain a pattern of having a `queue` prefix on all endpoints, which isn't perfect but it's better than nothing.  Do you have any other suggestions?  We could certainly fix this in the future in a major release.

omod/src/main/java/org/openmrs/module/queue/web/QueueEntryTransitionRestController.java
mseaton: Done.  I just find it uglier to read.  It would be happier if these optionals were not in the API.

api/src/main/java/org/openmrs/module/queue/model/QueueEntryTransition.java
mseaton: Yeah.  See the broader comment I made.

mseaton:
@ibacher - since you requested changes, I flagged you for a re-review.  I also had some followup questions/comments for a few things that you suggested.  I'm good to merge this as is and deal with some of these separately, but happy to continue to evolving if this needs more work

omod/src/main/java/org/openmrs/module/queue/web/QueueEntryTransitionRestController.java
ibacher: I see. Let's save it for a major release then.

omod/src/main/java/org/openmrs/module/queue/web/QueueEntryTransitionRestController.java
```diff
@@ -0,0 +1,109 @@
...
+import org.openmrs.module.webservices.rest.web.v1_0.controller.BaseRestController;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Controller;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.ResponseBody;
+
+/**
+ * The main controller that exposes additional end points for order entry
```
ojwanganto, line 34: This must have been an oversight. Should be something queues related

PR #59, opened 2024-02-27
O3-2904 Queue Module - queue should not have Queue Entries as a direc…

mseaton:
…t property

### DIFF
--- a/api/src/main/java/org/openmrs/module/queue/model/Queue.java
+++ b/api/src/main/java/org/openmrs/module/queue/model/Queue.java
@@ -22,18 +22,19 @@ import javax.persistence.OneToMany;
 import javax.persistence.Table;
 
 import java.util.ArrayList;
-import java.util.Date;
 import java.util.List;
 import java.util.stream.Collectors;
 
 import lombok.Data;
 import lombok.EqualsAndHashCode;
+import lombok.ToString;
 import org.apache.commons.lang.BooleanUtils;
 import org.openmrs.BaseChangeableOpenmrsMetadata;
 import org.openmrs.Concept;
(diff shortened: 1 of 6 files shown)


api/src/main/java/org/openmrs/module/queue/model/Queue.java
```diff
@@ -62,30 +63,9 @@ public class Queue extends BaseChangeableOpenmrsMetadata {
 	@JoinColumn(name = "status_concept_set", referencedColumnName = "concept_id")
 	private Concept statusConceptSet;
 	
-	@OneToMany(mappedBy = "queue", cascade = CascadeType.ALL, fetch = FetchType.LAZY)
-	private List<QueueEntry> queueEntries;
```
ibacher, line 66: I'm a little confused by this part of the change. Does this mean that queues no longer have queue entries?

mseaton: @ibacher yes, basically.  It says that "Queue Entries" belong to a Queue, but a Queue does not directly have access to the entries that are associated with it.  Can you find another example of where it's been our practice to have `data` belong to `metadata` directly?  You can't get a `List<PatientProgram>` from a `Program` for example, even if one might think they want to do something like `program.getEnrollments()`

ibacher: Nope. It always bothered me that we had this direct linkage! In retrospect, I should've just read the ticket.

ojwanganto: The refactoring makes sense. Thanks @mseaton

PR #60, opened 2024-03-05
O3-2861 - Queue Module - support default and configurable ordering of…

mseaton:
… queue entries on a queue

For a complete description of this change, please see most recent comments here:
https://openmrs.atlassian.net/browse/O3-2861

### DIFF
--- a/api/src/main/java/org/openmrs/module/queue/api/sort/BasicPrioritySortWeightGenerator.java
+++ b/api/src/main/java/org/openmrs/module/queue/api/sort/BasicPrioritySortWeightGenerator.java
@@ -1,0 +1,47 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.module.queue.api.sort;
+
+import java.util.List;
+
+import org.openmrs.Concept;
+import org.openmrs.module.queue.api.QueueServicesWrapper;
+import org.openmrs.module.queue.model.QueueEntry;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+/**
+ * Basic implementation of a sort weight generator that is based on the configured priority In this
+ * implementation, the position of the priority within the configured concept set determines the
+ * relative priority, where the set members are ordered in increasing priority. This means the lower
+ * the position in the concept set, the lower the sort weight (and the lower the priority)
+ */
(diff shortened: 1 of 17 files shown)


api/src/main/java/org/openmrs/module/queue/api/sort/BasicPrioritySortWeightGenerator.java
```diff
@@ -0,0 +1,47 @@
...
+	}
+	
+	/**
+	 * Generates a sort weight based on the order of the priority within the configured concept set
+	 */
+	public Double generateSortWeight(QueueEntry queueEntry) {
+		List<Concept> allowedPriorities = services.getAllowedPriorities(queueEntry.getQueue());
+		int ret = 0; // Default to the lowest sort weight
+		if (queueEntry.getPriority() != null && allowedPriorities.contains(queueEntry.getPriority())) {
+			ret = allowedPriorities.indexOf(queueEntry.getPriority());
```
mogoodrich, line 43: This seems backwards of how I'd expect a concept set to be sorted, I would figure the element that was sorted/displayed at the top would have index 0, not index length - 1, or am I reading this wrong.  (But not suggesting we change the way sort orders of concept sets work, just wanted to double-check this was correct!).

mogoodrich (approved):
LGTM

api/src/main/java/org/openmrs/module/queue/api/sort/BasicPrioritySortWeightGenerator.java
mseaton: You mean queues, not concept sets @mogoodrich ?  

For the set itself, we have tended to have the lower priority priorities listed first, as the default is often to use the first priority in the list, and we don't want the default to be an emergency.  And our PIH Priority Set has "Normal" then "Emergency".

For the queue, I tend to agree that a queue entries would more intuitively sort ascending.  And this is how I originally started implementing it.  But I changed it largely due to the fact that the existing / legacy queue ESM code had set "Emergency = 1" and "Normal = 0" in the database.  So if we want to keep in line with that, then we need to sort our sort orders descending.

mogoodrich: I meant concept sets, I guess my instinct would be that when viewing a concept set, the higher priority would be listed first, though based on what you said, this isn't the case.

So, what we are saying, is that if someone wanted to have a concept set "Low, Medium, HIgh" they'd have to arrange it (in the old concept set UI) like this:

Low
Medium
High

That just seems a little counter-intuitive, but maybe that's the standard.  The one example I found in CIEL does seem to follow that pattern:

https://app.openconceptlab.org/#/orgs/CIEL/sources/CIEL/concepts/161238/

mseaton:
@ibacher - I'm looking to merge but want to make sure you have a chance to weigh in if you have objections to this approach.

ojwanganto:
LGTM

api/src/main/java/org/openmrs/module/queue/api/sort/BasicPrioritySortWeightGenerator.java
ibacher: Either way, we should probably ensure this is documented somewhere.

ibacher (approved):

PR #61, opened 2024-03-08
O3-2942 - add more modifiable fields when editting/transitioning queue entries

chibongho:
Expand the list of updatable fields for queue entries. This enables the frontend of have a way to modify queue entries in place. Also changed the queue entry transition endpoint to support editing priorities comment.


mogoodrich (approved):
LGTM, thanks @chibongho !

mseaton (approved):
LGTM, thanks @chibongho !

PR #62, opened 2024-03-13
(fix):Filter Queue Entries to return count of those waiting for a ser…

makombe:
…vice

Essentially the queue metrics count should only those waiting for a particular service and not those who have finished with a particular service. The counts should also match what is displayed on the table.

Screenshot from 2024-03-13 17-18-07

### DIFF
--- a/omod/src/main/java/org/openmrs/module/queue/web/QueueEntryMetricRestController.java
+++ b/omod/src/main/java/org/openmrs/module/queue/web/QueueEntryMetricRestController.java
@@ -15,6 +15,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 import org.openmrs.module.queue.api.QueueServicesWrapper;
 import org.openmrs.module.queue.api.search.QueueEntrySearchCriteria;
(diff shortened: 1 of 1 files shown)


omod/src/main/java/org/openmrs/module/queue/web/QueueEntryMetricRestController.java
```diff
@@ -71,8 +72,12 @@ public Object handleRequest(HttpServletRequest request) {
 			ret.add(COUNT, services.getQueueEntryService().getCountOfQueueEntries(criteria).intValue());
 		} else {
 			List<QueueEntry> queueEntries = services.getQueueEntryService().getQueueEntries(criteria);
+			// Filter queueEntries where ended_at is null. The  count should be for those waiting for a service
+			List<QueueEntry> filteredQueueEntries = queueEntries.stream().filter(entry -> entry.getEndedAt() == null)
+			        .collect(Collectors.toList());
+
```
mseaton, line 78: @makombe thanks for this PR.  This endpoint does not make any assumptions about the metric being requested, and its intentionally kept generic.  As you can see, it will run a search based on the request parameters passed into it, and then return the relevant count of those search results.

I assume this is attempting to fix invalid metric data on the frontend, in which case the simple fix would be to pass an additional parameter in the frontend request of `isEnded=false`.  You can see the possible search criteria parameters in the `QueueEntrySearchCriteriaParser.java` class.

I recommend that we close this PR and instead make this fix in the service-queues app within patient managment.

mseaton (changes requested):
See inline comment.

makombe:
@mseaton I think that is ok. I will close it and handle it on the frontend.

PR #64, opened 2024-03-19
(BREAKING) O3-2748 add endpoint for un-doing transition of queue entries; ren…

chibongho:
…ame endpoints related to transitions

This PR add the ability to undo queue entry transitions. Undoing a transition voids the current active queue entry and un-ends its previous queue entry. 

I changed the route to REST endpoints related to queue entry transitions to be under `/queue-entry/*`, namely `/queue-entry/transition` and `queue-entry/undo-transition`, hence the breaking change.

### DIFF
--- a/api/src/main/java/org/openmrs/module/queue/api/QueueEntryService.java
+++ b/api/src/main/java/org/openmrs/module/queue/api/QueueEntryService.java
@@ -60,12 +60,15 @@ public interface QueueEntryService {
 	QueueEntry transitionQueueEntry(@NotNull QueueEntryTransition queueEntryTransition);
 	
 	/**
-	 * Undos a transition by voiding the input queue entry and making its previous queue entry 
-	 * (which MUST not be null) active by setting the previous entry's end time to null. 
+	 * Undos a transition to the input queue entry by voiding it and 
+   * making its previous queue entry active by
+	 * setting the previous entry's end time to null.
 	 * 
-	 * @param queueEntry the queue entry to undo transition from
+	 * @see QueueEntryService#getPreviousQueueEntry(QueueEntry)
+	 * @param queueEntry the queue entry to undo transition to. Must be active
 	 * @return the previous queue entry, re-activated
-	 * 
+	 * @throws IllegalArgumentException if the previous queue entry does not exist
+	 * @throws IllegalStateException if multiple previous entries are identified
 	 */
 	QueueEntry undoTransition(@NotNull QueueEntry queueEntry);
 	
@@ -125,11 +128,12 @@ public interface QueueEntryService {
 	void setSortWeightGenerator(SortWeightGenerator sortWeightGenerator);
 	
 	/**
-	 * Given a specified queue entry, return its previous queue entry (i.e. the queue entry the patient
-	 * transition from to get to the specified one)
+	 * Given a specified queue entry Q, return its previous queue entry P, where P has same patient and
+	 * visit as Q, and P.endedAt time is same as Q.startAt time.
 	 * 
 	 * @param queueEntry
-	 * @return the previous queue entry, if uniquely identifable; null otherwise.
+	 * @return the previous queue entry, null otherwise.
+	 * @throws IllegalStateException if multiple previous queue entries are identified
 	 */
 	QueueEntry getPreviousQueueEntry(@NotNull QueueEntry queueEntry);
 }
--- a/api/src/main/java/org/openmrs/module/queue/api/QueueEntryService.java
+++ b/api/src/main/java/org/openmrs/module/queue/api/QueueEntryService.java
@@ -60,9 +60,8 @@ public interface QueueEntryService {
 	QueueEntry transitionQueueEntry(@NotNull QueueEntryTransition queueEntryTransition);
 	
 	/**
-	 * Undos a transition to the input queue entry by voiding it and 
-   * making its previous queue entry active by
-	 * setting the previous entry's end time to null.
+	 * Undos a transition to the input queue entry by voiding it and making its previous queue entry
+	 * active by setting the previous entry's end time to null.
 	 * 
 	 * @see QueueEntryService#getPreviousQueueEntry(QueueEntry)
 	 * @param queueEntry the queue entry to undo transition to. Must be active
@@ -129,7 +128,8 @@ public interface QueueEntryService {
 	
 	/**
 	 * Given a specified queue entry Q, return its previous queue entry P, where P has same patient and
-	 * visit as Q, and P.endedAt time is same as Q.startAt time.
+	 * visit as Q, and P.endedAt time is same as Q.startedAt time, and P.queue is same as
+	 * Q.queueComingFrom
 	 * 
 	 * @param queueEntry
 	 * @return the previous queue entry, null otherwise.
--- a/api/src/main/java/org/openmrs/module/queue/api/QueueEntryService.java
+++ b/api/src/main/java/org/openmrs/module/queue/api/QueueEntryService.java
@@ -60,6 +60,16 @@ public interface QueueEntryService {
 	QueueEntry transitionQueueEntry(@NotNull QueueEntryTransition queueEntryTransition);
 	
 	/**
+	 * Undos a transition by voiding the input queue entry and making its previous queue entry 
+	 * (which MUST not be null) active by setting the previous entry's end time to null. 
+	 * 
+	 * @param queueEntry the queue entry to undo transition from
+	 * @return the previous queue entry, re-activated
+	 * 
+	 */
+	QueueEntry undoTransition(@NotNull QueueEntry queueEntry);
+	
+	/**
 	 * Voids a queue entry
 	 *
 	 * @param queueEntry the queue entry to be voided
@@ -113,4 +123,13 @@ public interface QueueEntryService {
 	 * @param sortWeightGenerator the SortWeightGenerator to set
 	 */
 	void setSortWeightGenerator(SortWeightGenerator sortWeightGenerator);
+	
+	/**
+	 * Given a specified queue entry, return its previous queue entry (i.e. the queue entry the patient
+	 * transition from to get to the specified one)
+	 * 
+	 * @param queueEntry
+	 * @return the previous queue entry, if uniquely identifable; null otherwise.
+	 */
+	QueueEntry getPreviousQueueEntry(@NotNull QueueEntry queueEntry);
 }
--- a/api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
+++ b/api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
@@ -17,6 +17,7 @@ import javax.validation.constraints.NotNull;
 import java.time.LocalDateTime;
 import java.time.LocalTime;
 import java.time.ZoneId;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Date;
 import java.util.List;
@@ -136,6 +137,8 @@ public class QueueEntryServiceImpl extends BaseOpenmrsService implements QueueEn
 	 */
 	@Override
 	public QueueEntry undoTransition(@NotNull QueueEntry queueEntry) {
+		// TODO: Exceptions should be translatable and human readable on the frontend.
+		// See: https://openmrs.atlassian.net/browse/O3-2988
 		if (queueEntry.getVoided()) {
 			throw new IllegalArgumentException("cannot undo transition on a voided queue entry");
 		}
@@ -257,10 +260,11 @@ public class QueueEntryServiceImpl extends BaseOpenmrsService implements QueueEn
 		criteria.setPatient(queueEntry.getPatient());
 		criteria.setVisit(queueEntry.getVisit());
 		criteria.setEndedOn(queueEntry.getStartedAt());
+		criteria.setQueues(Arrays.asList(queueEntry.getQueueComingFrom()));
 		List<QueueEntry> prevQueueEntries = dao.getQueueEntries(criteria);
 		if (prevQueueEntries.size() == 1) {
 			return prevQueueEntries.get(0);
-		} else if(prevQueueEntries.size() > 1) {
+		} else if (prevQueueEntries.size() > 1) {
 			throw new IllegalStateException("Multiple previous queue entries found");
 		} else {
 			return null;
--- a/api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
+++ b/api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
@@ -135,7 +135,6 @@ public class QueueEntryServiceImpl extends BaseOpenmrsService implements QueueEn
 	 * @see QueueEntryService#undoTransition(QueueEntry)
 	 */
 	@Override
-	@Transactional
 	public QueueEntry undoTransition(@NotNull QueueEntry queueEntry) {
 		if (queueEntry.getVoided()) {
 			throw new IllegalArgumentException("cannot undo transition on a voided queue entry");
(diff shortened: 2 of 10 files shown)


api/pom.xml
```diff
@@ -3,7 +3,7 @@
     <parent>
         <groupId>org.openmrs.module</groupId>
         <artifactId>queue</artifactId>
-        <version>2.3.0-SNAPSHOT</version>
+        <version>2.4.0-SNAPSHOT</version>
```
mseaton, line 6: Why is this getting changed from 2.3.0-SNAPSHOT -> 2.4.0-SNAPSHOT?  This would seem to indicate that this branch has not been updated with the latest from main.  Please merge main up to this branch so these changes no longer appear as within this PR.  I don't think we need to explicitly change the version in this PR.

api/src/main/java/org/openmrs/module/queue/api/QueueEntryService.java
```diff
@@ -59,6 +59,16 @@ public interface QueueEntryService {
 	 */
 	QueueEntry transitionQueueEntry(@NotNull QueueEntryTransition queueEntryTransition);
 	
+	/**
+	 * Undos a transition by voiding the input queue entry and making its previous queue entry 
+	 * (which MUST not be null) active by setting the previous entry's end time to null. 
+	 * 
+	 * @param queueEntry the queue entry to undo transition from
+	 * @return the previous queue entry, re-activated
```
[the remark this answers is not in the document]
mseaton: Agree the Javadocs can be a bit more expansive and explicit

api/src/main/java/org/openmrs/module/queue/api/QueueEntryService.java
```diff
@@ -113,4 +123,13 @@ String generateVisitQueueNumber(@NotNull Location location, @NotNull Queue queue
 	 * @param sortWeightGenerator the SortWeightGenerator to set
 	 */
 	void setSortWeightGenerator(SortWeightGenerator sortWeightGenerator);
+	
+	/**
+	 * Given a specified queue entry, return its previous queue entry (i.e. the queue entry the patient
+	 * transition from to get to the specified one)
```
mseaton, line 129: It isn't really an "i.e." condition.  There are specific rules and conditions here that we should document - specifically that the queue entries must have matching `queue`, `patient`, and `visit`, and the `endedAt` of the returned QueueEntry must match the `startedAt` of the passed QueueEntry exactly.

api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
```diff
@@ -131,6 +131,31 @@ public QueueEntry transitionQueueEntry(QueueEntryTransition queueEntryTransition
 		return getProxiedQueueEntryService().saveQueueEntry(queueEntryToStart);
 	}
 	
+	/**
+	 * @see QueueEntryService#undoTransition(QueueEntry)
+	 */
+	@Override
+	@Transactional
```
mseaton, line 138: The @Transactional annotation here is redundant, as it is added at the class level.

api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
```diff
@@ -225,4 +250,18 @@ private void endQueueEntry(@NotNull QueueEntry queueEntry) {
 		queueEntry.setEndedAt(new Date());
 		dao.createOrUpdate(queueEntry);
 	}
+	
+	@Override
+	public QueueEntry getPreviousQueueEntry(@NotNull QueueEntry queueEntry) {
```
mseaton, line 255: This needs to be marked as @Transactional(readOnly=true) to indicate this is a read-only operation

api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
```diff
@@ -225,4 +250,18 @@ private void endQueueEntry(@NotNull QueueEntry queueEntry) {
...
+		QueueEntrySearchCriteria criteria = new QueueEntrySearchCriteria();
+		criteria.setPatient(queueEntry.getPatient());
```
mseaton, line 257: The `Queue` also needs to match.  That appears to be missing here.

api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
```diff
@@ -225,4 +250,18 @@ private void endQueueEntry(@NotNull QueueEntry queueEntry) {
...
+		criteria.setVisit(queueEntry.getVisit());
+		criteria.setEndedOn(queueEntry.getStartedAt());
+		List<QueueEntry> prevQueueEntries = dao.getQueueEntries(criteria);
+		if (prevQueueEntries.size() == 1) {
+			return prevQueueEntries.get(0);
+		} else {
+			return null;
```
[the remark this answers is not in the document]
mseaton: I'd be inclined to throw an Exception here for both when there are 0 and > 1.

api/src/main/java/org/openmrs/module/queue/api/search/QueueEntrySearchCriteria.java
```diff
@@ -61,11 +61,15 @@ public class QueueEntrySearchCriteria implements Serializable {
...
 	
+	private Date startedOn;
+	
 	private Boolean isEnded = null;
 	
 	private Date endedOnOrAfter;
 	
 	private Date endedOnOrBefore;
 	
+	private Date endedOn;
```
mseaton, line 72: If we are expanding the capabilities of QueueEntrySearchCriteria, we should also add the necessary code to support these new search parameters via the REST endpoint.  Can you make the necessary improvements to `QueueEntrySearchCriteriaParser`?

integration-tests/pom.xml
```diff
@@ -3,7 +3,7 @@
     <parent>
         <groupId>org.openmrs.module</groupId>
         <artifactId>queue</artifactId>
-        <version>2.3.0-SNAPSHOT</version>
+        <version>2.4.0-SNAPSHOT</version>
```
mseaton, line 6: Per above, don't change

omod/pom.xml
```diff
@@ -3,7 +3,7 @@
     <parent>
         <groupId>org.openmrs.module</groupId>
         <artifactId>queue</artifactId>
-        <version>2.3.0-SNAPSHOT</version>
+        <version>2.4.0-SNAPSHOT</version>
```
mseaton, line 6: Per above, don't change

omod/src/main/java/org/openmrs/module/queue/web/QueueEntryTransitionRestController.java
```diff
@@ -34,80 +36,82 @@
...
 		
-		transition.setNewPriorityComment(body.get(NEW_PRIORITY_COMMENT));
+		transition.setNewPriorityComment(body.getNewPriorityComment());
 		
 		// Execute transition
 		QueueEntry newQueueEntry = services.getQueueEntryService().transitionQueueEntry(transition);
 		return ConversionUtil.convertToRepresentation(newQueueEntry, Representation.REF);
 	}
+	
+	@RequestMapping(value = "/rest/" + RestConstants.VERSION_1 + "/queue-entry/undo-transition", method = {
```
mseaton, line 104: I would think this would be better as a "DELETE" method on the `queue-entry/transition` endpoint, rather than a PUT/POST operation on a new endpoint.

omod/src/main/java/org/openmrs/module/queue/web/QueueEntryTransitionRestController.java
```diff
@@ -34,80 +36,82 @@
...
+	        RequestMethod.PUT, RequestMethod.POST })
+	@ResponseBody
+	public Object undoTransition(@RequestBody UndoQueueEntryTransitionRequest body) {
+		QueueEntryService qes = services.getQueueEntryService();
+		Optional<QueueEntry> queueEntry = qes.getQueueEntryByUuid(body.getQueueEntry());
+		if (queueEntry.isPresent()) {
+			QueueEntry unEndedQueueEntry = services.getQueueEntryService().undoTransition(queueEntry.get());
```
mseaton, line 111: Per above comments about throwing Exceptions, we should present these exceptions back to the user with appropriate error messages that can be adequately consumed in the UI

pom.xml
```diff
@@ -3,7 +3,7 @@
 
     <groupId>org.openmrs.module</groupId>
     <artifactId>queue</artifactId>
-    <version>2.3.0-SNAPSHOT</version>
+    <version>2.4.0-SNAPSHOT</version>
```
mseaton, line 6: Per above, remove this version change.

mseaton (changes requested):
Great job @chibongho .  See my review comments for some changes requested.

api/pom.xml
chibongho: The version bump is intentional. I made a backwards-incompatible change to the endpoint for transitions from `/queue-entry-transition` to `queue-entry/transition` (to namespace it better).

mogoodrich: Thanks @chibongho.  If we want to version bump, for Maven modules we can't just do a release via updated the version number... basically, we'd want to do a release OpenMRS CI (which would release 2.3.0 and then automatically bump the version to 2.4.0-SNAPSHOT) before merging this branch in.

But the other thing to check is that if there already are breaking changes in the 2.3.0-SNAPSHOT (ie since the 2.2.0 release) we won't need to do another release.

api/src/main/java/org/openmrs/module/queue/api/QueueEntryService.java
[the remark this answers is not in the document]
chibongho: > And just to verify own understanding basically when a patient is on a queue there will be a chain of queue entries representing their progress through the queue, and in order to "undo" an entry, that entry must be the "last" entry in the chain, and also not the first entries in the chain (ie there needs to be at least 2 entries)

@mogoodrich Yes, that is correct.

api/pom.xml
chibongho: Spoke to Mike. I did indeed mess up by local branch. main branch is already at 2.4 (and it hasn't been released yet) so i'll just keep that.

api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
chibongho: Wait... that's not true. You can transition to a different queue.

mseaton: Oh right, good point.

api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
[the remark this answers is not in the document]
mseaton: Looking at this again, if there are `0`, then it makes sense to return null.  So I think this is fine as is, with an Exception being thrown only for `> 1`.  For the message, we should either:
a) Make the message more informative for someone debugging, and indicate which specific queue entries were problematic, or
b) Make the message more consumable in a UI to present to a user (eg. return something that can be uniquely interpreted as a message key for a translated message.

api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
mseaton: Unresolving this.  You are right, you can transition to a different queue, but we do track the `queueComingFrom` on the new Queue Entry.  So in this case, we should set `criteria.setQueue(queueEntry.getQueueComingFrom())` if `queueEntry.getQueueComingFrom() != null`

api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
```diff
@@ -131,6 +131,30 @@ public QueueEntry transitionQueueEntry(QueueEntryTransition queueEntryTransition
...
+		if (queueEntry.getVoided()) {
+			throw new IllegalArgumentException("cannot undo transition on a voided queue entry");
+		}
+		if (queueEntry.getEndedAt() != null) {
+			throw new IllegalArgumentException("cannot undo transition on an ended queue entry");
+		}
+		QueueEntry prevQueueEntry = getPreviousQueueEntry(queueEntry);
+		if (prevQueueEntry == null) {
+			throw new IllegalArgumentException("specified queue entry does not have a previous queue entry");
+		}
```
mseaton, line 148: All 3 of the above exceptions would be best to throw in such a way that the UI can present them to the user in a translatable fashion, and one in which the message can be customized at the UI-level, not hard-coded on the backend level.  So it would be best to have some sort of code in the Exception that can be passed through to any error object returned to the frontend via the REST response.

mseaton (changes requested):
I resolved some comments that had been addressed, and added some additional @chibongho .  Ping me with the "Re-review" button when you think it's ready to go...

api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
```diff
@@ -225,4 +252,22 @@ private void endQueueEntry(@NotNull QueueEntry queueEntry) {
...
+		QueueEntrySearchCriteria criteria = new QueueEntrySearchCriteria();
+		criteria.setPatient(queueEntry.getPatient());
+		criteria.setVisit(queueEntry.getVisit());
+		criteria.setEndedOn(queueEntry.getStartedAt());
+		criteria.setQueues(Arrays.asList(queueEntry.getQueueComingFrom()));
+		List<QueueEntry> prevQueueEntries = dao.getQueueEntries(criteria);
+		if (prevQueueEntries.size() == 1) {
+			return prevQueueEntries.get(0);
+		} else if (prevQueueEntries.size() > 1) {
+			throw new IllegalStateException("Multiple previous queue entries found");
```
mseaton, line 268: Can you add the same TODO here regarding localized messages

mseaton (changes requested):
Just two quick things, and then I think this is ready.

api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
```diff
@@ -225,4 +252,25 @@ private void endQueueEntry(@NotNull QueueEntry queueEntry) {
...
+	
+	@Override
+	@Transactional(readOnly = true)
+	public QueueEntry getPreviousQueueEntry(@NotNull QueueEntry queueEntry) {
+		Queue queueComingFrom = queueEntry.getQueueComingFrom();
+		QueueEntrySearchCriteria criteria = new QueueEntrySearchCriteria();
+		criteria.setPatient(queueEntry.getPatient());
+		criteria.setVisit(queueEntry.getVisit());
+		criteria.setEndedOn(queueEntry.getStartedAt());
+		criteria.setQueues(queueComingFrom == null ? Arrays.asList() : Arrays.asList(queueComingFrom));
```
mseaton, line 264: What exactly do you want to happen here?  If you set this to an empty list, then the DAO will only return results in which the queue property of QueueEntry is null.  Since this is not (currently) nullable, this will effectively mean that no results are returned.  If that is _actually_ what we want, then we should just do a check earlier in the method and return null without bothering to hit the DB.  If it is _not_ what we want, and we want to get all queue entries that match regardless of queue (and if there is only one then use it), then we need the if statement to be around this entire critieria inclusion call.

mseaton (changes requested):
See comment.

mseaton (approved):

PR #69, opened 2024-04-11
(fix) O3-3066 update queue entries accordingly after merging patients

chibongho:
When merging 2 patients, all queue entries from the loser (voided) patient should be set to associate with the winner patient.

This PR does it by adding a SaveHandler on PersonMergeLog. Not sure if this is the right approach, since we have the same issue of needing to update queue entries when merging visits, we can't do something similar since there is no VisitMergeLog.


mseaton (commented):
One comment but generally looks good to me, assuming this works.  Do the save handlers operate in the same transaction?  Hard to test with our junit tests which work within a single transaction that gets rolled back, but I would think you could / should test what happens if there is an exception thrown by the SaveHandler - do things end up in an invalid state?

chibongho:
Closing this as this approach does not work, unfortunately. The SaveHandler does not seem to be called within the same transaction that initiates the merge.

PR #65, opened 2024-03-28
O3-3002: Queue Module - REST endpoints can be accessed without authentication.

IamMujuziMoses:
## Issue I worked on

see https://issues.openmrs.org/browse/O3-3002

## Checklist: I completed these to help reviewers :)

1. Added `Authorized` annotations to the service classes: `QueueService`, `QueueEntryService`, `QueueRoomService` and `RoomProviderMapService`.
2. Created `PrivilegeConstants` class to contain all privilege names and their descriptions.

PR #70, opened 2024-04-17
Revert "O3-3002: Queue Module - REST endpoints can be accessed without authentication."

mseaton:
Reverts openmrs/openmrs-module-queue#65

@dkayiwa / @IamMujuziMoses - unfortunately this commit appears to be breaking our implementation due to the new privileges required.  I can't currently access JIRA due to the infrastructure outage to see any past discussion here, so I'm going to revert it for now so we can discuss and figure out the best way to get these improvements committed.


dkayiwa:
@mseaton what would be your recommended approach to address the issues that you are facing with your implementation?

mseaton:
@dkayiwa - I made some suggestions just now on the ticket - lemme know what you think.

PR #74, opened 2024-05-09
(feat) O3-3184 allow changing of LocationWaitingFor and ProviderWaiti…

chibongho:
…ngFor during a transition

This PR add `locationWaitingFor` and `providerWaitingFor` as attributes that are changeable during a transition. 

The logic here is a bit strange since these fields may be set to null, and we need to way to distinguish between "changing the field to null" vs "keeping the field as it was". It does that by determining whether the optional attributes `newLocationWaitingFor` and `newProviderWaitingFor` exists in the `/queue-entry/transition` request body. If not, those fields do not get changed.

### DIFF
--- a/api/src/main/java/org/openmrs/module/queue/model/QueueEntryTransition.java
+++ b/api/src/main/java/org/openmrs/module/queue/model/QueueEntryTransition.java
@@ -14,6 +14,8 @@ import java.util.Date;
 
 import lombok.Data;
 import org.openmrs.Concept;
+import org.openmrs.Location;
+import org.openmrs.Provider;
 
 /**
  * Bean definition that encapsulates the supported criteria for saving a direct transition from one
@@ -36,6 +38,10 @@ public class QueueEntryTransition implements Serializable {
 	
 	private String newPriorityComment;
 	
+	private Location newLocationWaitingFor; // value obtrained from queueEntryToTransition if not provided by request
+	
+	private Provider newProviderWaitingFor; // value obtrained from queueEntryToTransition if not provided by request
+	
 	/**
 	 * @return a new queue entry representing what one intends to transition into
 	 */
@@ -49,8 +55,8 @@ public class QueueEntryTransition implements Serializable {
 		    newPriorityComment == null ? queueEntryToTransition.getPriorityComment() : newPriorityComment);
 		queueEntry.setStatus(newStatus == null ? queueEntryToTransition.getStatus() : newStatus);
 		queueEntry.setSortWeight(queueEntryToTransition.getSortWeight());
-		queueEntry.setLocationWaitingFor(queueEntryToTransition.getLocationWaitingFor());
-		queueEntry.setProviderWaitingFor(queueEntryToTransition.getProviderWaitingFor());
+		queueEntry.setLocationWaitingFor(newLocationWaitingFor);
+		queueEntry.setProviderWaitingFor(newProviderWaitingFor);
 		queueEntry.setQueueComingFrom(queueEntryToTransition.getQueue());
 		queueEntry.setStartedAt(transitionDate);
 		return queueEntry;
--- a/omod/src/main/java/org/openmrs/module/queue/web/QueueEntryTransitionRestController.java
+++ b/omod/src/main/java/org/openmrs/module/queue/web/QueueEntryTransitionRestController.java
@@ -10,9 +10,13 @@
 package org.openmrs.module.queue.web;
 
 import java.util.Date;
+import java.util.Map;
 import java.util.Optional;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
 import org.openmrs.Concept;
+import org.openmrs.Location;
+import org.openmrs.Provider;
 import org.openmrs.api.APIException;
 import org.openmrs.module.queue.api.QueueEntryService;
 import org.openmrs.module.queue.api.QueueServicesWrapper;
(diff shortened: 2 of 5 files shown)


omod/src/main/java/org/openmrs/module/queue/web/QueueEntryTransitionRestController.java
```diff
@@ -96,6 +101,36 @@ public Object transitionQueueEntry(@RequestBody QueueEntryTransitionRequest body
 		
 		transition.setNewPriorityComment(body.getNewPriorityComment());
 		
+		// Location waiting for
+		if (rawBody.containsKey(QueueEntryTransitionRequest.NEW_LOCATION_WAITING_FOR_FIELD)) {
```
ibacher, line 105: Was there a reason these couldn't just be added as properties for the QueueEntryTransitionRequest?

ibacher (approved):
Seems fine. One question, though.

omod/src/main/java/org/openmrs/module/queue/web/QueueEntryTransitionRestController.java
chibongho: They are properties already, but having those properties be null is not sufficient to distinguish between "we intend to set this to null" vs "we don't want to change this property". So we check the rawBody to see if those optional properties are provided at all, and only set the corresponding queue entry field if so.

api/src/main/java/org/openmrs/module/queue/model/QueueEntryTransition.java
```diff
@@ -36,6 +38,10 @@ public class QueueEntryTransition implements Serializable {
 	
 	private String newPriorityComment;
 	
+	private Location newLocationWaitingFor; // value obtrained from queueEntryToTransition if not provided by request
+	
+	private Provider newProviderWaitingFor; // value obtrained from queueEntryToTransition if not provided by request
```
mseaton, line 43: Typo in both comments "obtrained" -> "obtained"

omod/src/main/java/org/openmrs/module/queue/web/QueueEntryTransitionRestController.java
mseaton: @chibongho - isn't that also the case for some of the other, already existing properties you are setting?  For example `priorityComment`?

api/src/main/java/org/openmrs/module/queue/model/QueueEntryTransition.java
```diff
@@ -49,8 +55,8 @@ public QueueEntry constructNewQueueEntry() {
 		    newPriorityComment == null ? queueEntryToTransition.getPriorityComment() : newPriorityComment);
 		queueEntry.setStatus(newStatus == null ? queueEntryToTransition.getStatus() : newStatus);
 		queueEntry.setSortWeight(queueEntryToTransition.getSortWeight());
-		queueEntry.setLocationWaitingFor(queueEntryToTransition.getLocationWaitingFor());
-		queueEntry.setProviderWaitingFor(queueEntryToTransition.getProviderWaitingFor());
+		queueEntry.setLocationWaitingFor(newLocationWaitingFor);
+		queueEntry.setProviderWaitingFor(newProviderWaitingFor);
```
mseaton, line 59: It seems a little inconsistent here to do this vs. what we do for status, priority, etc.  For the existing fields, we handle the logic here for setting them to existing fields.  For these new properties, we handle that logic in the REST controller.  Would it make sense to be consistent?

omod/src/main/java/org/openmrs/module/queue/web/QueueEntryTransitionRestController.java
```diff
@@ -96,6 +101,36 @@ public Object transitionQueueEntry(@RequestBody QueueEntryTransitionRequest body
...
+			if (body.getNewLocationWaitingFor() == null) {
+				transition.setNewLocationWaitingFor(null);
```
mseaton, line 107: This seems unnecessary.  Isn't it always null already in this case?

omod/src/main/java/org/openmrs/module/queue/web/QueueEntryTransitionRestController.java
```diff
@@ -96,6 +101,36 @@ public Object transitionQueueEntry(@RequestBody QueueEntryTransitionRequest body
...
+				transition.setNewLocationWaitingFor(location);
+			}
+		} else {
+			transition.setNewLocationWaitingFor(queueEntry.getLocationWaitingFor());
+		}
+		
+		// Provider waiting for
+		if (rawBody.containsKey(QueueEntryTransitionRequest.NEW_PROVIDER_WAITING_FOR_FIELD)) {
+			if (body.getNewProviderWaitingFor() == null) {
+				transition.setNewProviderWaitingFor(null);
```
mseaton, line 122: Same comment as above.

omod/src/main/java/org/openmrs/module/queue/web/QueueEntryTransitionRestController.java
```diff
@@ -96,6 +101,36 @@ public Object transitionQueueEntry(@RequestBody QueueEntryTransitionRequest body
...
+			} else {
+				Provider provider = services.getProviderService().getProviderByUuid(body.getNewLocationWaitingFor());
```
mseaton, line 124: This is a bug, probably related to copying the code block.  This should be `body.getNewProviderWaitingFor()`.  I'm surprised the tests didn't catch this - can you modify the tests to ensure they do?

mseaton (changes requested):

omod/src/main/java/org/openmrs/module/queue/web/QueueEntryTransitionRestController.java
chibongho: @mseaton that's a good point about `priorityComment` (and other nullable fields). We got away with it because the frontend defaults the `QueueEntryTransitionRequest.newPriorityComment` field to whatever the original queue entry's `priorityComment` is, or empty string if its null. The UI renders null or empty string the same way anyway.

I originally intended the `QueueEntryTransitionRequest` to have all fields be optional and the request would only modify the fields that are provided. That seems to not work as well for nullable fields. I can just change the API so that all fields are required, and the request would have to populate all non-modified fields with values from the original queue entry. (We already do that in the fronend).

chibongho:
Closing this PR for now due to ticket reprioritization.

PR #80, opened 2025-02-19
(fix) O3-4416 void queue entries when corresponding visit voids

chibongho:


api/src/main/java/org/openmrs/module/queue/api/VisitWithQueueEntriesSaveHandler.java
```diff
@@ -51,5 +52,18 @@ public void handle(Visit visit, User user, Date date, String s) {
...
 			}
 		}
+		if (visit.getVisitId() != null && visit.getVoided()) {
+			QueueEntrySearchCriteria criteria = new QueueEntrySearchCriteria();
+			criteria.setVisit(visit);
+			List<QueueEntry> queueEntries = queueEntryService.getQueueEntries(criteria);
+			for (QueueEntry qe : queueEntries) {
+				if (!qe.getVoided()) {
+					qe.setVoided(true);
+					qe.setVoidReason("visit voided");
```
mseaton, line 62: Probably set this to the same void reason as the visit.voidReason, as well as set the "dateVoided" and voidedBy to match the values set on the Visit dateVoided and voidedBy.

mseaton (approved):
LGTM, assuming this works :)

PR #94, opened 2026-02-23
O3-5445: Fix duplicate SQL query in AbstractBaseQueueDaoImpl#get(String uuid)

sanks011:
## Description
`AbstractBaseQueueDaoImpl#get(String uuid)` was calling `criteria.add(eq("uuid", uuid)).uniqueResult()` twice on the same mutable `Criteria` object - firing two SQL queries per UUID lookup and adding a redundant duplicate `WHERE uuid = ?` clause on the second query.

This affected every UUID-based entity lookup across the module (`QueueEntry`, `Queue`, `QueueRoom`, `RoomProviderMap`).

Fix removes the orphaned first call and keeps a single `criteria.add(...) → uniqueResult()` chain.

## Related
https://openmrs.atlassian.net/jira/software/c/projects/O3/boards/37?selectedIssue=O3-5445


bhavya-jpg:
@sanks011 
Ah, that makes perfect sense! Thank you so much for taking the time to explain the difference in how SessionStatistics.getEntityLoadCount() tracks those invocations. That's a really good nuance to know regarding uniqueResult() calls!

It's also great to hear that all the concrete implementations inherit properly from AbstractBaseQueueDaoImpl so this single fix handles everything downstream natively without us needing to run an audit. I appreciate the architectural context.

bhavya-jpg (approved):

PR #102, opened 2026-03-15
O3-5485 Fix NaN average wait time when all queue entries have null endedAt

RoshanMadival08:
## Description
Fixes an issue where computeAverageWaitTimeInMinutes could return NaN when all queue entries had null endedAt values.

## Changes
- Updated the logic to only calculate the average using entries that have both startedAt and endedAt.
- Added a check to return 0 when there are no valid entries to calculate the wait time.
- Added unit tests to cover this case.

## Testing
Ran the test suite using:
mvn clean test

All tests passed successfully.

## Related
https://openmrs.atlassian.net/browse/O3-5485


sourav-jyoti:
hi @RoshanMadival08 , theres already a pending pr for this tckt, from next time raise pr only if theres no pending pr

PR #109, opened 2026-04-07
(fix) O3-5594:  Fix errors when voiding patient with multiple queue e…

chibongho:
…ntries

This PR fixes the following errors when voiding patient with multiple queue entries:
- `dao.getOverlappingQueueEntries()` function does not take into account of endAt time correctly in the overlap check
- the `isDuplicate()` function check fails because it loops through voided queue entries. Even though the `getOverlappingQueueEntries()` call within the function should only return non-voided entries, it doesn’t actually do that for entries being voided during a cascade void (of the patient) because the void value hasn’t been committed to the DB yet.


mseaton (approved):
LGTM

PR #113, opened 2026-05-21
O3-5682: Ship Visit Queue Number visit attribute type

UjjawalPrabhat:
## Summary
Adds a Liquibase changeset that inserts the `visitQueueNumber` VisitAttributeType (UUID `c0c579b0-8e59-401d-8a4a-976a0b183519`) on module install/upgrade. Idempotent via `sqlCheck` precondition.

### Why?
The queue screen (`/spa/home/service-queues/screen`) is blank in default installs because the Call→serve flow depends on a "Visit Queue Number" visit attribute that nothing in the module ever creates. The frontend has long had a `visitQueueNumberAttributeUuid` config knob that was meant to be overridden by each distribution, but in practice the Reference Application and most installs ran on the `null` default — so `generateVisitQueueNumber` silently no-ops, visits never get a ticket number, and the assignticket POST silently 200s with nothing stored. Shipping the VAT in the module itself gives every install a working default; distributions that want their own UUID can still override via config.

## Related Issue 
[O3-5682](https://openmrs.atlassian.net/browse/O3-5682)

[O3-5682]: https://openmrs.atlassian.net/browse/O3-5682?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/src/main/resources/liquibase.xml
+++ b/api/src/main/resources/liquibase.xml
@@ -766,7 +766,7 @@
             <column name="min_occurs" valueNumeric="0"/>
             <column name="max_occurs" valueNumeric="1"/>
             <column name="creator" valueNumeric="1"/>
-            <column name="date_created" valueDate="NOW()"/>
+            <column name="date_created" valueComputed="NOW()"/>
             <column name="retired" valueBoolean="false"/>
             <column name="uuid" value="c0c579b0-8e59-401d-8a4a-976a0b183519"/>
         </insert>
--- a/api/src/main/resources/liquibase.xml
+++ b/api/src/main/resources/liquibase.xml
@@ -749,4 +749,27 @@
         </sql>
     </changeSet>
 
+    <changeSet id="add_visit_queue_number_attribute_type_20260521" author="ujjawalprabhat">
+        <preConditions onFail="MARK_RAN">
+            <sqlCheck expectedResult="0">
+                SELECT COUNT(*) FROM visit_attribute_type WHERE uuid = 'c0c579b0-8e59-401d-8a4a-976a0b183519';
+            </sqlCheck>
+        </preConditions>
+        <comment>
+            Insert the "Visit Queue Number" visit attribute type used by the queue screen to display ticket numbers.
+            UUID matches the default value of `visitQueueNumberAttributeUuid` in @openmrs/esm-service-queues-app.
+        </comment>
+        <insert tableName="visit_attribute_type">
+            <column name="name" value="Visit Queue Number"/>
+            <column name="description" value="The visit queue number used to identify patients on the queue screen"/>
+            <column name="datatype" value="org.openmrs.customdatatype.datatype.FreeTextDatatype"/>
+            <column name="min_occurs" valueNumeric="0"/>
+            <column name="max_occurs" valueNumeric="1"/>
+            <column name="creator" valueNumeric="1"/>
+            <column name="date_created" valueDate="NOW()"/>
+            <column name="retired" valueBoolean="false"/>
+            <column name="uuid" value="c0c579b0-8e59-401d-8a4a-976a0b183519"/>
+        </insert>
+    </changeSet>
+
 </databaseChangeLog>
--- a/api/src/main/java/org/openmrs/module/queue/api/digitalSignage/QueueTicketAssignments.java
+++ b/api/src/main/java/org/openmrs/module/queue/api/digitalSignage/QueueTicketAssignments.java
@@ -11,6 +11,7 @@ package org.openmrs.module.queue.api.digitalSignage;
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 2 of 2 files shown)


api/src/main/resources/liquibase.xml
```diff
@@ -749,4 +749,27 @@
...
+            UUID matches the default value of `visitQueueNumberAttributeUuid` in @openmrs/esm-service-queues-app.
+        </comment>
+        <insert tableName="visit_attribute_type">
+            <column name="name" value="Visit Queue Number"/>
+            <column name="description" value="The visit queue number used to identify patients on the queue screen"/>
+            <column name="datatype" value="org.openmrs.customdatatype.datatype.FreeTextDatatype"/>
+            <column name="min_occurs" valueNumeric="0"/>
+            <column name="max_occurs" valueNumeric="1"/>
+            <column name="creator" valueNumeric="1"/>
+            <column name="date_created" valueDate="NOW()"/>
```
jwnasambu: @UjjawalPrabhat Thanks for putting this fix together. A minor recommendation please! kindly change valueDate="NOW()" to valueComputed="NOW()". This explicitly tells Liquibase to evaluate the function natively regardless of the target database.

**Note:**  While MySQL database  might gracefully accept NOW(), PostgreSQL which is heavily used in OpenMRS environments will throw an error because NOW() is an unquoted function name treated as a literal string value by default in standard  tags. 
Reference: https://docs.liquibase.com/secure/reference-guide-5-1-1/change-types/column
 Feel free to follow the discussion on this link : https://forum.liquibase.org/t/liquibase-3-3-insert-sysdate-into-date-field-fails-with-ora-01858/3142

jwnasambu (commented):
@UjjawalPrabhat Kindly I want to believe  with the current changes the screen remains blank because it expects a results array but the API returns a raw Map. 

  2.  Don't we need to add a patientName field to the assignments model?
  
 3. Kindly for safety use a ConcurrentHashMap for the in memory storage to prevent server crashes during simultaneous polling and updates.
 
Thanks.

UjjawalPrabhat:
> Kindly I want to believe with the current changes the screen remains blank because it expects a results array but the API returns a raw Map.

I had some concerns before going on with this one:

- Breaking API change. Anything consuming `GET /queueutil/active-tickets` today (frontend, integration tests, third-party clients) gets the raw `Map<String, TicketAssignment>`. Switching to `{results: [...]}` breaks that contract silently.
- Inconsistent with the rest of the controller. The other utility endpoints in `Legacy1xRestController (/queue-entry-number, /queueutil/metrics-by-service)` also return custom shapes — the `{results: [...]}` envelope is for `BaseDelegatingResource`-backed domain resources, not ad-hoc utility endpoints. Changing only this one makes it the odd one out.
- Doesn't necessarily fixes [O3-5682](https://openmrs.atlassian.net/browse/O3-5682). Screen renders correctly with the current shape once the VAT exists — verified locally. Probably belongs in a separate "make `queueutil` endpoints OMRS-conventional" ticket if there's appetite for that refactor. 

Happy to defer if you feel strongly.

> 2. Don't we need to add a patientName field to the assignments model?

it wasn't present before and even if we need it, should it be included in this PR?
> 3. Kindly for safety use a ConcurrentHashMap for the in memory storage to prevent server crashes during simultaneous polling and updates.

done.



[O3-5682]: https://openmrs.atlassian.net/browse/O3-5682?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

jwnasambu (commented):
@UjjawalPrabhat Kindly doesn't the PR title need to reference the ticket ID?  https://openmrs.atlassian.net/wiki/spaces/docs/pages/25477199/Pull+Request+Tips

UjjawalPrabhat:
> @UjjawalPrabhat Kindly doesn't the PR title need to reference the ticket ID? 

Isn't O3-5682 the ticketID?

api/src/main/resources/liquibase.xml
```diff
@@ -749,4 +749,27 @@
...
     </changeSet>
 
+    <changeSet id="add_visit_queue_number_attribute_type_20260521" author="ujjawalprabhat">
+        <preConditions onFail="MARK_RAN">
+            <sqlCheck expectedResult="0">
+                SELECT COUNT(*) FROM visit_attribute_type WHERE uuid = 'c0c579b0-8e59-401d-8a4a-976a0b183519';
+            </sqlCheck>
+        </preConditions>
+        <comment>
+            Insert the "Visit Queue Number" visit attribute type used by the queue screen to display ticket numbers.
```
NethmiRodrigo, line 759: Ah, we don’t usually insert things like this into the database because these are things that the implementers need to configure. Instead, we use content packages. If this is a visit attribute type, then it should go into a configuration file like this one - https://github.com/openmrs/openmrs-content-referenceapplication-demo/blob/main/configuration/backend_configuration/attributetypes/attribute_types-core_demo.csv. These get processed by the [initializer](https://github.com/mekomsolutions/openmrs-module-initializer/blob/main/readme/atttypes.md) and saved to the database

jwnasambu:
@UjjawalPrabhat Kindly this is how your ticket Id should look link `O3-5682: Call/Serve action flow is broken: queue-screen stays blank` and not `O3-5682: Ship Visit Queue Number visit attribute type`

NethmiRodrigo:
Closing this since - https://github.com/openmrs/openmrs-module-queue/pull/113#discussion_r3299164057

PR #114, opened 2026-05-25
O3-5692: Build a dedicated REST endpoint for queue entries

UjjawalPrabhat:
## Summary
The Service Queues table currently loads via `QueueEntryResource` with a deep custom `v=` representation that pulls every encounter, obs, and diagnosis on each patient's visit. On a 50-entry queue this triggers hundreds of cascading Hibernate loads; ~95% of the bytes are never read by the table.

Adds `GET /rest/v1/queue-entry-summary` — flat, server-paginated, with `previousQueueEntryUuid` resolved via a single batch IN-clause query instead of the per-row N+1 path through `getPreviousQueueEntry`.

Mirrors the three-layer batch pattern from [openmrs-module-emrapi PR #250](https://github.com/openmrs/openmrs-module-emrapi/pull/250).

### Response shape

**Before** (`/queue-entry?v=custom:(…)` as requested by the frontend):

```json
{
  "uuid": "...",
  "display": "...",
  "queue":    { "uuid": "...", "display": "...", "name": "...", "description": "...", "service": {...}, "location": {...} },
  "status":   { "uuid": "...", "display": "...", "name": {...}, "datatype": {...}, "conceptClass": {...}, "set": false, "version": "...", "retired": false, "names": [...], "descriptions": [...], "mappings": [...], "answers": [...], "setMembers": [...], ... },
  "priority": { /* same deep Concept shape */ },
  "patient": {
    "uuid": "...", "display": "...",
    "identifiers": [ { "uuid": "...", "display": "...", "identifier": "...", "identifierType": {...} } ],
    "person": { /* full person graph */ }
  },
  "visit": {
    "uuid": "...", "startDatetime": "...",
    "encounters": [
      {
        "uuid": "...", "display": "...", "encounterDatetime": "...",
        "encounterType": {...},
        "diagnoses": [ { /* full diagnosis */ } ],
        "obs": [ { /* full obs, recursively */ } ],
        "encounterProviders": [ { /* full provider */ } ],
        "voided": false
      }
      /* …every encounter on the visit, each with its full obs/diagnoses/providers graph */
    ],
    "attributes": [ { /* full visit attributes */ } ]
  },
  "priorityComment": "...",
  "sortWeight": 0.0,
  "startedAt": "...", "endedAt": null,
  "locationWaitingFor": {...}, "providerWaitingFor": {...}, "queueComingFrom": {...},
  "previousQueueEntry": { /* QueueEntry at default rep; its nested previousQueueEntry is a REF stub, so it stops there — not unbounded */ }

}
```

After (`/queue-entry-summary`):
```json
{
  "uuid": "...",
  "patient":  { "uuid": "...", "display": "100GEJ - John Doe" },
  "queue":    { "uuid": "...", "display": "Triage" },
  "status":   { "uuid": "...", "display": "Waiting" },
  "priority": { "uuid": "...", "display": "Not urgent" },
  "priorityComment": "...",
  "startedAt": "2026-05-25T09:00:00.000+0000",
  "visitUuid": "...",
  "visitStartDatetime": "2026-05-25T08:55:00.000+0000",
  "previousQueueEntryUuid": null
}
```

## Relate Issue
[O3-5692](https://openmrs.atlassian.net/browse/O3-5692)

[O3-5692]: https://openmrs.atlassian.net/browse/O3-5692?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jwnasambu (commented):
@UjjawalPrabhat this is a good progress . Kindly make the following updates here:

- Update SUMMARY_REPRESENTATION to include patientIdentifier.
- Add logic to rename patientIdentifier to identifier and flatten the patient object.

omod/src/main/java/org/openmrs/module/queue/web/QueueEntrySummaryRestController.java
```diff
@@ -0,0 +1,97 @@
...
+		
+		QueueEntryService queueEntryService = services.getQueueEntryService();
+		List<QueueEntry> entries = queueEntryService.getQueueEntries(criteria, context.getStartIndex(), context.getLimit());
+		Map<QueueEntry, String> previousUuids = queueEntryService.getPreviousQueueEntryUuids(entries);
+		
+		CustomRepresentation entryRep = new CustomRepresentation(SUMMARY_REPRESENTATION);
+		List<SimpleObject> results = new ArrayList<>(entries.size());
+		for (QueueEntry entry : entries) {
+			SimpleObject row = (SimpleObject) ConversionUtil.convertToRepresentation(entry, entryRep);
+			SimpleObject visit = (SimpleObject) row.remove("visit");
```
jwnasambu, line 81: Kindly add this logic in between line 76 and 77
```suggestion
		 SimpleObject patient = (SimpleObject) row.get("patient");
          if (patient != null) {
              patient.add("identifier", patient.remove("patientIdentifier"));
      }
```
What this loop does is to wait for ConversionUtil.convertToRepresentation to finish so the row has patient data to modify.Then group all "flat shape" changes at the top of the loop before appending to the results list.

omod/src/main/java/org/openmrs/module/queue/web/QueueEntrySummaryRestController.java
```diff
@@ -0,0 +1,97 @@
...
+ * Lightweight, flat-shape, server-paginated read endpoint for the Service Queues table. Returns
+ * only the fields the queue table renders; avoids the deep nested encounter/obs/diagnosis graph
+ * that the default {@code QueueEntryResource} representation pulls in. Three-layer batch-load
+ * pattern mirrors openmrs-module-emrapi PR #250.
+ */
+@Controller
+public class QueueEntrySummaryRestController extends BaseRestController {
+	
+	private static final String SUMMARY_REPRESENTATION = "uuid,patient:(uuid,display),queue:(uuid,display),"
+	        + "status:(uuid,display),priority:(uuid,display),priorityComment,startedAt,visit:(uuid,startDatetime)";
```
jwnasambu:
```suggestion
	 private static final String SUMMARY_REPRESENTATION = "uuid,patient:(uuid,display,patientIdentifier),queue:(uuid,display),"
    "status:(uuid,display),priority:(uuid,display),priorityComment,startedAt,visit:(uuid,startDatetime)";

```

jwnasambu (commented):
Also create this file to include test cases for
     getQueueEntrySummaries_shouldReturnPaginatedSummariesWithIdentifier,
     shouldHandleTotalCountRequest, and shouldReturnEmptyListWhenNoResults. I know the PR originally focuses on DAO tests but without Controller tests there is
     no verification that the final JSON keys like identifier or visitUuid are correctly
     transformed for the frontend.

api/src/main/java/org/openmrs/module/queue/api/QueueEntryService.java
```diff
@@ -129,6 +131,20 @@ public interface QueueEntryService {
...
+	 */
+	@Authorized({ PrivilegeConstants.GET_QUEUE_ENTRIES })
+	List<QueueEntry> getQueueEntries(@NotNull QueueEntrySearchCriteria searchCriteria, Integer startIndex, Integer limit);
+	
+	/**
+	 * Batch-resolves the uuid of the previous queue entry for each input entry that has a non-null
+	 * {@code queueComingFrom}. Entries without a resolvable predecessor are absent from the result.
+	 */
+	@Authorized({ PrivilegeConstants.GET_QUEUE_ENTRIES })
+	Map<QueueEntry, String> getPreviousQueueEntryUuids(Collection<QueueEntry> entries);
```
jwnasambu, line 146: Kindly to follow the 3-layer architectural pattern (Controller -> Service -> DAO) mentioned in the PR description, please expose
  this method `getPreviousQueueEntryUuids`  in the Service interface to ensures that the logic is available to the REST
  controller while maintaining proper security and transaction boundaries. please don't forget to add the @Authorized annotation for
  PrivilegeConstants.GET_QUEUE_ENTRIES."

api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
```diff
@@ -215,6 +217,18 @@ public List<QueueEntry> getQueueEntries(QueueEntrySearchCriteria searchCriteria)
...
+	@Transactional(readOnly = true)
+	public List<QueueEntry> getQueueEntries(QueueEntrySearchCriteria searchCriteria, Integer startIndex, Integer limit) {
+		return dao.getQueueEntries(searchCriteria, startIndex, limit);
+	}
+	
+	@Override
+	@Transactional(readOnly = true)
+	public Map<QueueEntry, String> getPreviousQueueEntryUuids(Collection<QueueEntry> entries) {
+		return dao.getPreviousQueueEntryUuids(entries);
+	}
```
jwnasambu, line 230: Kindly  provide the implementation for `getPreviousQueueEntryUuids` by delegating the call to the DAO layer. This maintains
  consistency with the patterns used in emrapi and ensures the service layer remains the primary entry point for business logic.

omod/src/main/java/org/openmrs/module/queue/web/QueueEntrySummaryRestController.java
```diff
@@ -0,0 +1,97 @@
...
+	@RequestMapping(value = "/rest/" + RestConstants.VERSION_1 + "/queue-entry-summary", method = RequestMethod.GET)
+	public ResponseEntity<?> getQueueEntrySummaries(HttpServletRequest request, HttpServletResponse response) {
+		@SuppressWarnings("unchecked")
+		Map<String, String[]> parameterMap = request.getParameterMap();
+		QueueEntrySearchCriteria criteria = parser.constructFromRequest(parameterMap);
+		RequestContext context = RestUtil.getRequestContext(request, response, Representation.DEFAULT);
+		
+		QueueEntryService queueEntryService = services.getQueueEntryService();
+		List<QueueEntry> entries = queueEntryService.getQueueEntries(criteria, context.getStartIndex(), context.getLimit());
+		Map<QueueEntry, String> previousUuids = queueEntryService.getPreviousQueueEntryUuids(entries);
```
jwnasambu, line 71: In the controller, please ensure you are calling queueEntryService.getPreviousQueueEntryUuids() rather than trying to
  access the DAO directly. This preserves the architectural integrity of the module.

api/src/main/java/org/openmrs/module/queue/api/QueueEntryService.java
UjjawalPrabhat: This is already in place, the method is declared on `QueueEntryService` at line 146 with `@Authorized({ PrivilegeConstants.GET_QUEUE_ENTRIES })` on line 145. Let me know if you're seeing something different.

api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
UjjawalPrabhat: The implementation at lines 228-230 delegates to the DAO - `return dao.getPreviousQueueEntryUuids(entries);` with `@Override` + `@Transactional(readOnly = true)` on the lines above. Happy to adjust if you'd like a different pattern.

omod/src/main/java/org/openmrs/module/queue/web/QueueEntrySummaryRestController.java
UjjawalPrabhat: The controller calls the service, not the DAO — `queueEntryService.getPreviousQueueEntryUuids(entries)` on line 71, where `queueEntryService` is obtained from `services.getQueueEntryService()` on line 69. There's no DAO reference anywhere in the controller.

chibongho:
Hi @UjjawalPrabhat, some questions I have:

- Are you sure that `visit.encounter.obs` and `previousQueueEntry` are pulled in recursively? Looking that [this](https://github.com/openmrs/openmrs-esm-patient-management/blob/fe29ef139895de91f97912276f77e801831ea89d/packages/esm-service-queues-app/src/hooks/useQueueEntries.ts#L8) from the O3 service queues app, I don't think we are doing that.
- I'm not surprised that the frontend is pulling more data than is needed. That said, can't we fix that by fixing the `custom:()` rep in the frontend?
- You're right that `previousQueueEntry` is not that efficient. If `/queueEntry?custom(previousQueueEntry)` returns N queue entries, then for each N of them, we make one additional request for its previous queue entry. It looks like this will help reduce it to one query instead. That said, is it that a big performance improvement to warrant having a specialized endpoint?
  - `preivousQueueEntry` is unfortunately not well implemented, see [code](https://github.com/openmrs/openmrs-module-queue/blob/1bbdfe9919c4a949735eb21687301d856a727eae/api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java). The right way to do it would have been having an actual `preivous_queue_entry` column in the DB. That way, we could have added the following to [QueueEntry.java](https://github.com/openmrs/openmrs-module-queue/blob/ba5242b8550dbfb3d5363e1a135469c38fde390f/api/src/main/java/org/openmrs/module/queue/model/QueueEntry.java), which would fix the performance issue:
```
        @BatchSize(size = 100) // this will automatically batch the queries for previousQueueEntries for a List<QueueEntry> returned by hibernate
        @Column(name = "previous_queue_entry")
        private QueueEntry previousQueueEntry;
```

UjjawalPrabhat:
> * Are you sure that `visit.encounter.obs` and `previousQueueEntry` are pulled in recursively? Looking that [this](https://github.com/openmrs/openmrs-esm-patient-management/blob/fe29ef139895de91f97912276f77e801831ea89d/packages/esm-service-queues-app/src/hooks/useQueueEntries.ts#L8) from the O3 service queues app, I don't think we are doing that.

The table's rep is the inline one in [useQueueEntries.ts:9](https://github.com/openmrs/openmrs-esm-patient-management/blob/fe29ef139895de91f97912276f77e801831ea89d/packages/esm-service-queues-app/src/constants.ts#L12) (not [constants.ts](https://github.com/openmrs/openmrs-esm-patient-management/blob/fe29ef139895de91f97912276f77e801831ea89d/packages/esm-service-queues-app/src/constants.ts#L12)), identical at the commit you linked. It does request `encounters:(...,obs,diagnoses,encounterProviders,...)`, all bare → default reps, so obs/diagnoses/providers are pulled per entry. You're right about `previousQueueEntry` though, it's only one level (nested previous is REF, [QueueEntryResource:248](https://github.com/openmrs/openmrs-module-queue/blob/main/omod/src/main/java/org/openmrs/module/queue/web/resources/QueueEntryResource.java#L248)) and not recursive.

> * I'm not surprised that the frontend is pulling more data than is needed. That said, can't we fix that by fixing the `custom:()` rep in the frontend?

Mostly yes. The encounter/obs/diagnosis graph, `visit.attributes`, `person`, and `locationWaitingFor`/`providerWaitingFor` are unused by the table and can be dropped from the rep with no backend change (they're lazy, so it cuts the load too). 

What a rep-trim won't do: add DB-level pagination or batch `previousQueueEntry` (And the table uses `useOpenmrsFetchAll`, which pulls all pages. So DB pagination only helps if the table moves to true pagination).

> * You're right that `previousQueueEntry` is not that efficient. If `/queueEntry?custom(previousQueueEntry)` returns N queue entries, then for each N of them, we make one additional request for its previous queue entry. It looks like this will help reduce it to one query instead. That said, is it that a big performance improvement to warrant having a specialized endpoint?

Agreed the N+1 is real. The `previous_queue_entry` column fixes it for all consumers - cost is a migration + backfill + write-path change (and it'd be `@ManyToOne @JoinColumn`, not `@Column`, for `@BatchSize`). 

One catch either way: the summary returns only `previousQueueEntryUuid`, but the frontend reads `previousQueueEntry.startedAt` and `.status.display`, so the uuid alone won't cover today's usage.

@ibacher @NethmiRodrigo Any thoughts??

ibacher:
@UjjawalPrabhat I think I agree with Chi Bong's overall point here that maybe we should try just making these changes via the frontend and see how far that gets us. Pagination is, I think, not something we should be worried about too much. By design, the queue should be a relatively reasonable size (i.e., < 100 active entries). I think a custom backend makes sense if we needed to preserve elements of the full object graph (because while the custom representations are nice, they are not GraphQL, so we can't just say "grab obs with these 6 concepts we care about"). If we're just dropping those completely, let's see how far this can get without a specialized REST endpoint.

UjjawalPrabhat:
@chibongho @ibacher New PRs' raised [here](https://github.com/openmrs/openmrs-esm-patient-management/pull/2583) and [here](https://github.com/openmrs/openmrs-module-queue/pull/117).

CC: @NethmiRodrigo @denniskigen

PR #115, opened 2026-06-03
O3-5696: Fix Wrong ORM mapping, duplicate DB lookup & null crash

UjjawalPrabhat:
Three independent backend defects in the queue module:

- `QueueEntry.queueComingFrom` was mapped `@OneToOne` though many entries can transition from the same source queue. Switch to `@ManyToOne`, matching the other associations. Mapping correctness only, no behaviour change: Hibernate binds an owning-side `@OneToOne` with a join column as a `ManyToOneType` anyway, and the generated schema is identical.

- `AbstractBaseQueueDaoImpl.get(uuid)` executed its criteria query twice, discarding the first result. Drop the redundant call so each UUID lookup runs exactly one query (affects all queue DAOs via the base).

- `getDisplay` null-checked `personName` but dereferenced a possibly-null patient first, in both `QueueEntryResource` and `QueueEntrySubResource`. Null-check the patient and fall back to the queue entry UUID, in both, with unit tests. Hardening rather than a reproducible crash: `patient_id` is NOT NULL and validation rejects a null patient, so no in-module path writes such a row. It guards rows written from outside, and makes the two getters agree.

### DIFF
--- a/api/src/main/java/org/openmrs/module/queue/model/QueueEntry.java
+++ b/api/src/main/java/org/openmrs/module/queue/model/QueueEntry.java
@@ -16,7 +16,6 @@ import javax.persistence.GenerationType;
 import javax.persistence.Id;
 import javax.persistence.JoinColumn;
 import javax.persistence.ManyToOne;
-import javax.persistence.OneToOne;
 import javax.persistence.Table;
 
 import java.util.Date;
@@ -89,7 +88,7 @@ public class QueueEntry extends BaseChangeableOpenmrsData {
 	
 	//The queue the patient is coming from, if any.
 	@ToString.Exclude
-	@OneToOne
+	@ManyToOne
 	@JoinColumn(name = "queue_coming_from", referencedColumnName = "queue_id")
 	private Queue queueComingFrom;
 	
--- a/omod/src/main/java/org/openmrs/module/queue/web/resources/QueueEntryResource.java
+++ b/omod/src/main/java/org/openmrs/module/queue/web/resources/QueueEntryResource.java
@@ -25,6 +25,7 @@ import io.swagger.models.properties.RefProperty;
 import io.swagger.models.properties.StringProperty;
 import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
+import org.openmrs.Patient;
 import org.openmrs.PersonName;
 import org.openmrs.api.context.Context;
 import org.openmrs.module.queue.api.QueueServicesWrapper;
@@ -281,8 +282,12 @@ public class QueueEntryResource extends DelegatingCrudResource<QueueEntry> {
 	
 	@PropertyGetter("display")
 	public String getDisplay(QueueEntry queueEntry) {
-		PersonName personName = queueEntry.getPatient().getPersonName();
-		return (personName == null ? queueEntry.getPatient().toString() : personName.getFullName());
+		Patient patient = queueEntry.getPatient();
+		if (patient == null) {
+			return queueEntry.getUuid();
+		}
+		PersonName personName = patient.getPersonName();
+		return (personName == null ? patient.toString() : personName.getFullName());
 	}
 	
 	@PropertyGetter("previousQueueEntry")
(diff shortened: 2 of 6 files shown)


jwnasambu:
@claude review

omod/src/main/java/org/openmrs/module/queue/web/resources/QueueEntryResource.java
```diff
@@ -281,8 +282,12 @@ private void addSharedResourceDescriptionProperties(DelegatingResourceDescriptio
 	
 	@PropertyGetter("display")
 	public String getDisplay(QueueEntry queueEntry) {
-		PersonName personName = queueEntry.getPatient().getPersonName();
-		return (personName == null ? queueEntry.getPatient().toString() : personName.getFullName());
+		Patient patient = queueEntry.getPatient();
+		if (patient == null) {
+			return queueEntry.getUuid();
+		}
```
dkayiwa, line 288: `QueueEntrySubResource.getDisplay` needs the same guard: one file over, at `omod/src/main/java/org/openmrs/module/queue/web/resources/QueueEntrySubResource.java:235`, the identical dereference is still unguarded, and there it happens twice.

```java
if (queueEntry.getPatient().getPerson().getPersonName() == null) {
	return "";
}
return queueEntry.getPatient().getPerson().getPersonName().getFullName();
```

Same entity, same `display` property, served from `GET /ws/rest/v1/queue/{uuid}/entry`, so a row that would have crashed the getter you just fixed still takes that endpoint down. Fixing only one of the two leaves half the exposure in place.

One thing I couldn't pin down: what actually produces a queue entry with a null patient. `patient_id` is NOT NULL in liquibase.xml, and on create the getter never sees an unvalidated delegate, since `DelegatingCrudResource.create` runs `ValidateUtil.validate` before `save` and before the conversion, and `QueueEntryValidator` rejects a null patient (I saved one through the proxied service to check, and got `ValidationException ... patient: queueEntry.patient.null`). The only `setPatient` on a queue entry copies it off the entry being transitioned. So this reads as hardening against rows written outside the module, which is a fine thing to do, I just wouldn't harden one getter and not its twin.

UjjawalPrabhat: Fixed, the sub-resource getter now mirrors the one in `QueueEntryResource`. Note the null-`personName` branch changed from `""` to `patient.toString()` so the two agree.

On the open question: I couldn't produce a null patient either. Three layers reject it - through the proxied service, `ValidationException` (core's `RequiredDataAdvice` validates `saveQueueEntry`, not the impl itself); with validation disabled, Hibernate `PropertyValueException`; raw SQL insert, `NULL not allowed for column "PATIENT_ID"`. So it's hardening against rows written outside the module.

One correction: `DelegatingSubResource` never calls `ValidateUtil` at all, in `create` or `update`, so validates-before-conversion doesn't cover this getter, only the service AOP does. And the frontend does reach it: requeue and `endPatientStatus` both POST to `/queue/{uuid}/entry/{uuid}`, which converts at `DEFAULT`, where `display` lives.

api/src/main/java/org/openmrs/module/queue/model/QueueEntry.java
```diff
@@ -89,7 +88,7 @@ public class QueueEntry extends BaseChangeableOpenmrsData {
 	
 	//The queue the patient is coming from, if any.
 	@ToString.Exclude
-	@OneToOne
+	@ManyToOne
```
dkayiwa, line 91: Did the `@OneToOne` actually bite somewhere, or did it come out of reading the annotation? I went looking for the constraint it implies and couldn't find one on 2.7.4 (Hibernate 5.6.15). Hibernate binds an owning-side `@OneToOne` with a join column as a `ManyToOneType` either way, the only difference being an internal `isLogicalOneToOne` flag; the generated H2 schema has no unique key on `queue_coming_from`, just the FK; and the SQL emitted for loading a queue entry is byte identical before and after. I also saved a second entry coming from queue 1 while entry 2 in the test dataset already had `queue_coming_from=1`, and it inserted fine on main. Production schema comes from liquibase rather than from the annotations, and that only adds the FK too.

None of which is an argument against the change, the annotation should say what's true and I'd keep it. But it isn't fixing a defect, and the description ("implying each source queue belongs to a single entry") reads like it is. Since that text is what lands in the history, could you reword that bullet as a mapping correction with no behavior change? Happy to be corrected if you saw it fail on a platform version other than the one this builds against.

UjjawalPrabhat: You're right, and it came from reading the annotation rather than from a failure. I reproduced your result: dumping H2's `INFORMATION_SCHEMA` for `queue_entry` under both annotations gives identical output for `queue_coming_from`, same FK, `NON_UNIQUE=TRUE`, no unique key either way. Both also default to EAGER, so the fetch plan is unchanged too.

Reworded the bullet as a mapping correction with no behaviour change.

dkayiwa:
@claude review

PR #116, opened 2026-06-05
O3-5697: Add indexes to queue_entry for queue_id, patient_id & started_at/ended_at

UjjawalPrabhat:
## Summary
The `queue_entry` table of Service Queues  has no indexes on the columns those queries filter by (`queue_id`, `patient_id`, `started_at`, `ended_at`). Every query therefore scans the entire table.

On a fresh demo install with a handful of patients this is invisible. But in a real clinic with months or years of accumulated queue history (potentially tens of thousands of rows), the queue page gets progressively slower until staff notice it taking seconds to load and by then the table is large enough that adding indexes needs careful migration planning. Indexing it now, while it's small, avoids that.

## Related Issue
[O3-5697](https://openmrs.atlassian.net/browse/O3-5697)

[O3-5697]: https://openmrs.atlassian.net/browse/O3-5697?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/src/main/resources/liquibase.xml
+++ b/api/src/main/resources/liquibase.xml
@@ -749,4 +749,25 @@
         </sql>
     </changeSet>
 
+    <changeSet id="add_queue_entry_indexes_20260605" author="ujjawalprabhat">
+        <preConditions onFail="MARK_RAN" onError="WARN">
+            <not>
+                <indexExists tableName="queue_entry" indexName="queue_entry_started_ended_idx"/>
+            </not>
+        </preConditions>
+        <comment>
+            Add indexes to queue_entry (queue_id, patient_id, and composite started_at/ended_at)
+        </comment>
+        <createIndex tableName="queue_entry" indexName="queue_entry_queue_id_idx">
+            <column name="queue_id"/>
+        </createIndex>
+        <createIndex tableName="queue_entry" indexName="queue_entry_started_ended_idx">
+            <column name="started_at"/>
+            <column name="ended_at"/>
+        </createIndex>
+        <createIndex tableName="queue_entry" indexName="queue_entry_patient_id_idx">
+            <column name="patient_id"/>
+        </createIndex>
+    </changeSet>
+
 </databaseChangeLog>


jwnasambu (commented):
Thanks, @UjjawalPrabhat  good and focused change.

jwnasambu (approved):

api/src/main/resources/liquibase.xml
```diff
@@ -749,4 +749,25 @@
         </sql>
     </changeSet>
 
+    <changeSet id="add_queue_entry_indexes_20260605" author="ujjawalprabhat">
+        <preConditions onFail="MARK_RAN" onError="WARN">
+            <not>
```
NethmiRodrigo, line 754: This precondition only checks one of the three indexes, and MySQL commits each `CREATE INDEX` on its own. Let's split this into three changesets, one index each, each with its own `indexExists` precondition, like the other changesets in this file.

api/src/main/resources/liquibase.xml
```diff
@@ -749,4 +749,25 @@
...
+        <comment>
+            Add indexes to queue_entry (queue_id, patient_id, and composite started_at/ended_at)
+        </comment>
+        <createIndex tableName="queue_entry" indexName="queue_entry_queue_id_idx">
+            <column name="queue_id"/>
+        </createIndex>
+        <createIndex tableName="queue_entry" indexName="queue_entry_started_ended_idx">
+            <column name="started_at"/>
+            <column name="ended_at"/>
+        </createIndex>
```
NethmiRodrigo: The queries that actually get slow filter on `ended_at IS NULL` (the REST layer even defaults `isEnded` to `false`), and an index that leads with `started_at` can't serve that. The `(queue_id, ended_at)` one also covers the FK on `queue_id`, so we don't need a separate `queue_id` index.

```suggestion
        <createIndex tableName="queue_entry" indexName="queue_entry_queue_ended_idx">
             <column name="queue_id"/>
             <column name="ended_at"/>
         </createIndex>
         <createIndex tableName="queue_entry" indexName="queue_entry_ended_at_idx">
             <column name="ended_at"/>
         </createIndex>
         <createIndex tableName="queue_entry" indexName="queue_entry_started_at_idx">
             <column name="started_at"/>
         </createIndex>
```

api/src/main/resources/liquibase.xml
```diff
@@ -749,4 +749,53 @@
...
+        <comment>Add index to queue_entry on started_at</comment>
+        <createIndex tableName="queue_entry" indexName="queue_entry_started_at_idx">
+            <column name="started_at"/>
+        </createIndex>
+    </changeSet>
+
+    <changeSet id="add_queue_entry_patient_id_index_20260605" author="ujjawalprabhat">
+        <preConditions onFail="MARK_RAN" onError="WARN">
+            <not>
+                <indexExists tableName="queue_entry" indexName="queue_entry_patient_id_idx"/>
```
NethmiRodrigo: The `queue_entry_patient_id_fk` FK already indexes this column, so this would create a duplicate index on existing installs
````suggestion
                <indexExists tableName="queue_entry" columnNames="patient_id"/>
````

NethmiRodrigo (commented):
Thanks @UjjawalPrabhat! Just one more thing -

NethmiRodrigo (approved):
Thanks @UjjawalPrabhat!

PR #117, opened 2026-06-13
O3-5693: Resolve previousQueueEntry via a column to avoid N+1

UjjawalPrabhat:
## Summary
The queue REST page resolved `previousQueueEntry` with one query per row (`getPreviousQueueEntry` ran a criteria lookup per entry), an N+1 on every load.

This persists the relationship instead: a new `previous_queue_entry` column on `queue_entry`, set when the entry is created — directly for a transition, otherwise derived from `queueComingFrom` — and read straight off the column thereafter. `QueueEntry` carries `@BatchSize(size = 100)`, so the predecessors for a page of entries load in one round trip rather than one query each. Existing rows are backfilled where the predecessor is unambiguous (MySQL/MariaDB and Postgres changesets).

## Behaviour changes
- An ambiguous predecessor now resolves to null. Previously `getPreviousQueueEntry` threw `IllegalStateException("Multiple previous queue entries found")` on read — during list serialization, so it could fail a whole page render.
- Pre-upgrade rows rely on the backfill, so a database outside MySQL/MariaDB/Postgres keeps null links for existing entries. New writes are unaffected.
- Undoing a transition clears the voided entry's link rather than leaving it pointing at the re-opened predecessor.
- Where an entry has no visit, the runtime lookup doesn't constrain on visit (unchanged from the old criteria) while the backfill requires both sides to be null.
- The FK uses `ON DELETE SET NULL`, unlike the others in `liquibase.xml`: `purgeQueueEntry` hard-deletes, and `NO ACTION` would make purging any entry that has a successor fail.

## Related Issue
[O3-5693](https://openmrs.atlassian.net/browse/O3-5693)

## Other
Related PRs' [here](https://github.com/openmrs/openmrs-module-queue/pull/114) and [here](https://github.com/openmrs/openmrs-esm-patient-management/pull/2583).

[O3-5693]: https://openmrs.atlassian.net/browse/O3-5693?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/src/main/java/org/openmrs/module/queue/api/QueueEntryService.java
+++ b/api/src/main/java/org/openmrs/module/queue/api/QueueEntryService.java
@@ -68,8 +68,8 @@ public interface QueueEntryService {
 	 * time is the same as Q.startedAt time, and whose queue is Q.queueComingFrom
 	 *
 	 * @param queueEntry
-	 * @return the previous queue entry, null if there is none or if it could not be identified
-	 *         unambiguously when the queue entry was created.
+	 * @return the previous queue entry, null if there is none, if it could not be identified
+	 *         unambiguously when the queue entry was created, or if it has since been voided.
 	 */
 	@Authorized(PrivilegeConstants.GET_QUEUE_ENTRIES)
 	QueueEntry getPreviousQueueEntry(@NotNull QueueEntry queueEntry);
--- a/api/src/main/java/org/openmrs/module/queue/api/QueueEntryService.java
+++ b/api/src/main/java/org/openmrs/module/queue/api/QueueEntryService.java
@@ -62,13 +62,14 @@ public interface QueueEntryService {
 	List<QueueEntry> getOverlappingQueueEntries(Patient patient, Queue queue, Date startedAt, Date endedAt);
 	
 	/**
-	 * Given a specified queue entry Q, return its previous queue entry P, where P has same patient and
-	 * visit as Q, and P.endedAt time is same as Q.startedAt time, and P.queue is same as
-	 * Q.queueComingFrom
+	 * Given a specified queue entry Q, return the previous queue entry recorded against it. This is set
+	 * when Q is created, either directly by a transition or, for a queue entry saved with a
+	 * queueComingFrom, by looking up the entry P with the same patient and visit as Q, whose endedAt
+	 * time is the same as Q.startedAt time, and whose queue is Q.queueComingFrom
 	 *
 	 * @param queueEntry
-	 * @return the previous queue entry, null otherwise.
-	 * @throws IllegalStateException if multiple previous queue entries are identified
+	 * @return the previous queue entry, null if there is none or if it could not be identified
+	 *         unambiguously when the queue entry was created.
 	 */
 	@Authorized(PrivilegeConstants.GET_QUEUE_ENTRIES)
 	QueueEntry getPreviousQueueEntry(@NotNull QueueEntry queueEntry);
@@ -100,7 +101,6 @@ public interface QueueEntryService {
 	 * @param queueEntry the queue entry to undo transition to. Must be active
 	 * @return the previous queue entry, re-activated
 	 * @throws IllegalArgumentException if the previous queue entry does not exist
-	 * @throws IllegalStateException if multiple previous entries are identified
 	 */
 	@Authorized({ PrivilegeConstants.MANAGE_QUEUE_ENTRIES })
 	QueueEntry undoTransition(@NotNull QueueEntry queueEntry);
--- a/api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
+++ b/api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
@@ -134,6 +134,7 @@ public class QueueEntryServiceImpl extends BaseOpenmrsService implements QueueEn
 		queueEntryTransition.setTransitionDate(transitionDate);
 		
 		QueueEntry queueEntryToStart = queueEntryTransition.constructNewQueueEntry();
+		queueEntryToStart.setPreviousQueueEntry(queueEntryToStop);
 		
 		// Use optimistic locking to end the current entry
 		queueEntryToStop.setEndedAt(transitionDate);
@@ -295,27 +296,6 @@ public class QueueEntryServiceImpl extends BaseOpenmrsService implements QueueEn
 	@Override
 	@Transactional(readOnly = true)
 	public QueueEntry getPreviousQueueEntry(@NotNull QueueEntry queueEntry) {
-		Queue queueComingFrom = queueEntry.getQueueComingFrom();
-		if (queueComingFrom == null) {
-			return null;
-		}
-		
-		QueueEntrySearchCriteria criteria = new QueueEntrySearchCriteria();
-		criteria.setPatient(queueEntry.getPatient());
-		criteria.setVisit(queueEntry.getVisit());
-		criteria.setEndedOn(queueEntry.getStartedAt());
-		criteria.setQueues(Collections.singletonList(queueComingFrom));
-		
-		List<QueueEntry> prevQueueEntries = dao.getQueueEntries(criteria);
-		
-		if (prevQueueEntries.size() == 1) {
-			return prevQueueEntries.get(0);
-		} else if (prevQueueEntries.size() > 1) {
-			// TODO: Exceptions should be translatable and human readable on the frontend.
-			// See: https://openmrs.atlassian.net/browse/O3-2988
-			throw new IllegalStateException("Multiple previous queue entries found");
-		} else {
-			return null;
-		}
+		return queueEntry.getPreviousQueueEntry();
 	}
 }
--- a/api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
+++ b/api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
@@ -300,10 +300,13 @@ public class QueueEntryServiceImpl extends BaseOpenmrsService implements QueueEn
 	@Override
 	@Transactional(readOnly = true)
 	public QueueEntry getPreviousQueueEntry(@NotNull QueueEntry queueEntry) {
-		return queueEntry.getPreviousQueueEntry();
+		QueueEntry previousQueueEntry = queueEntry.getPreviousQueueEntry();
+		if (previousQueueEntry == null || previousQueueEntry.getVoided()) {
+			return null;
+		}
+		return previousQueueEntry;
 	}
 	
-	// Resolves the predecessor from queueComingFrom, returning null if there is no unambiguous match.
 	private QueueEntry resolvePreviousQueueEntry(QueueEntry queueEntry) {
 		Queue queueComingFrom = queueEntry.getQueueComingFrom();
 		if (queueComingFrom == null) {
--- a/api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
+++ b/api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
@@ -93,6 +93,9 @@ public class QueueEntryServiceImpl extends BaseOpenmrsService implements QueueEn
 	public QueueEntry saveQueueEntry(QueueEntry queueEntry) {
 		Double sortWeight = getSortWeightGenerator().generateSortWeight(queueEntry);
 		queueEntry.setSortWeight(sortWeight);
+		if (queueEntry.getId() == null && queueEntry.getPreviousQueueEntry() == null) {
+			queueEntry.setPreviousQueueEntry(resolvePreviousQueueEntry(queueEntry));
+		}
 		return dao.createOrUpdate(queueEntry);
 	}
 	
@@ -134,7 +137,6 @@ public class QueueEntryServiceImpl extends BaseOpenmrsService implements QueueEn
 		queueEntryTransition.setTransitionDate(transitionDate);
 		
 		QueueEntry queueEntryToStart = queueEntryTransition.constructNewQueueEntry();
-		queueEntryToStart.setPreviousQueueEntry(queueEntryToStop);
 		
 		// Use optimistic locking to end the current entry
 		queueEntryToStop.setEndedAt(transitionDate);
@@ -184,6 +186,8 @@ public class QueueEntryServiceImpl extends BaseOpenmrsService implements QueueEn
 			throw new IllegalStateException("Previous queue entry was modified by another transaction");
 		}
 		
+		// Cleared before voiding so that voidQueueEntry's own save persists it, rather than a second write
+		queueEntry.setPreviousQueueEntry(null);
 		getProxiedQueueEntryService().voidQueueEntry(queueEntry, "Transition undone");
 		
 		// Reload the previous entry to return the updated state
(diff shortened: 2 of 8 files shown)


api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
```diff
@@ -134,6 +134,7 @@ public QueueEntry transitionQueueEntry(QueueEntryTransition queueEntryTransition
 		queueEntryTransition.setTransitionDate(transitionDate);
 		
 		QueueEntry queueEntryToStart = queueEntryTransition.constructNewQueueEntry();
+		queueEntryToStart.setPreviousQueueEntry(queueEntryToStop);
```
jwnasambu: Nit: Kindly you added `queueEntryToStart.setPreviousQueueEntry(queueEntryToStop)` here but the existing `shouldUndoTransitionQueueEntry` test has no assertion verifying the field is set. The integration test covers it, but the unit test should pin this contract so regressions are caught independently.

api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
jwnasambu, line 191: Kindly this PR replaces the on-the-fly `getPreviousQueueEntry` computation with a persisted previous_queue_entry column. When undoTransition voids the entry, the FK is left intact, so a voided row still claims a predecessor. The old code had nothing to clear because the relationship was computed from non-voided entries only. The new column changes that contract. 


```suggestion
		getProxiedQueueEntryService().voidQueueEntry(queueEntry, "Transition undone");
queueEntry.setPreviousQueueEntry(null);
dao.createOrUpdate(queueEntry);
```

this keeps the data model consistent with the prior implicit behavior: a voided entry should not have a predecessor reference.

dkayiwa:
@claude review

api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
```diff
@@ -295,27 +296,6 @@ private static Date roundToSecond(Date date) {
...
-		if (prevQueueEntries.size() == 1) {
-			return prevQueueEntries.get(0);
-		} else if (prevQueueEntries.size() > 1) {
-			// TODO: Exceptions should be translatable and human readable on the frontend.
-			// See: https://openmrs.atlassian.net/browse/O3-2988
-			throw new IllegalStateException("Multiple previous queue entries found");
-		} else {
-			return null;
-		}
+		return queueEntry.getPreviousQueueEntry();
```
dkayiwa: Now that resolution reads only the persisted column, `getPreviousQueueEntry` returns non-null only for entries created through `transitionQueueEntry` (plus historical rows the backfill managed to link). A couple of cases that used to resolve now come back null:

- an entry created by POSTing to `/queue-entry` with `queueComingFrom` set (still a creatable property on this resource): the old criteria lookup would have found the predecessor, the column never gets set;
- rows the backfill skips: ambiguous ones, and any database other than MySQL/MariaDB/Postgres. In the test dataset entry 2's `queue_coming_from` lines up with entry 1, so on H2 `getPreviousQueueEntry(entry2)` now returns null where it used to return entry 1. The new null-case test sidesteps this by picking entry 3 instead.

I traced the O3 flow and it transitions through `/queue-entry/transition`, so the frontend itself is fine. Mainly want to confirm the narrowing to transition-only is intended. If it is, it may be worth dropping `queueComingFrom` from the creatable properties (or at least noting the change), so a client that sets it doesn't silently get a null previous entry. If it isn't, falling back to the old computation when the column is null would keep it working.

api/src/main/resources/liquibase.xml
```diff
@@ -749,4 +749,71 @@
...
+        <comment>
+            Backfill previous_queue_entry for existing transitioned entries, only where the predecessor
+            (same patient, same visit, queue = queue_coming_from, ended_at = started_at, not voided) is unambiguous.
+        </comment>
+        <sql>
+            update queue_entry curr
+            inner join (
+                select curr2.queue_entry_id as curr_id, min(prev.queue_entry_id) as prev_id, count(*) as n
+                from queue_entry curr2
+                inner join queue_entry prev
```
dkayiwa, line 835: One edge in this join: `prev` can match `curr` itself. If a row has `ended_at = started_at` (possible for older entries created before the strictly-positive-duration rule) and `queue_id = queue_coming_from` (exactly what a same-queue transition produces), it satisfies its own predecessor conditions, so with no other candidate it backfills `previous_queue_entry` to its own id. Adding `and prev.queue_entry_id <> curr2.queue_entry_id` to the join rules that out. Same thing applies to the Postgres changeset just below.

Worth being careful here because neither backfill is exercised by the tests: both are gated to `mysql,mariadb` / `postgresql`, so H2 skips them and the SQL first runs for real during a production migration. Running it once against a representative MySQL copy before release would be a good safety net.

api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
dkayiwa: I went looking for a concrete failure from the leftover FK and didn't find one in the current tree: nothing reverse-queries `previous_queue_entry`, and the list queries exclude voided entries, so the stale reference on a voided row never gets read or shown. So I'd treat clearing it as optional cleanup rather than something that blocks.

If you do clear it, one small thing on the snippet: `voidQueueEntry` already calls `dao.createOrUpdate(queueEntry)` internally, so the extra `createOrUpdate` after it is a second write to the same row. Setting `queueEntry.setPreviousQueueEntry(null)` before the `voidQueueEntry` call persists it in the same save.

api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
UjjawalPrabhat: Narrowing wasn't intended, so instead of a read-time fallback the old computation now runs once at write time in `saveQueueEntry`, a POST with `queueComingFrom` still links and the property stays creatable. A fallback would keep re-querying on every page load for rows whose column is legitimately null.

Noting the two behaviour changes, as you asked: ambiguity now returns null instead of throwing (the old `IllegalStateException` fired during list serialization), and pre-upgrade rows depend on the backfill, so databases outside MySQL/MariaDB/Postgres keep null links.

On the H2 case: I couldn't link entry 2 in the shared dataset, it breaks the two `QueueEntryDaoTest` purge tests, since they purge entry 1 and H2's hbm2ddl FK has no `ON DELETE SET NULL`. Covered with a transition → flush/clear → re-read test instead.

jwnasambu (approved):

jwnasambu:
@dkayiwa, @ibacher though I have approved the PR I would appreciate another pair of eyes before merge.

api/src/main/resources/liquibase.xml
```diff
@@ -749,4 +749,67 @@
...
+            <columnExists tableName="queue_entry" columnName="previous_queue_entry"/>
+        </preConditions>
+        <comment>Backfill previous_queue_entry for existing entries whose predecessor is unambiguous</comment>
+        <sql>
+            update queue_entry curr
+            inner join (
+                select curr2.queue_entry_id as curr_id, min(prev.queue_entry_id) as prev_id, count(*) as n
+                from queue_entry curr2
+                inner join queue_entry prev
+                    on prev.queue_entry_id <> curr2.queue_entry_id
```
dkayiwa: This has to be escaped before merge. `<>` puts a raw `<` into element content, so liquibase.xml is no longer well-formed XML and Liquibase rejects the whole changelog before it reaches any changeset. Same operator at line 802 in the Postgres changeset below.

Running the parser Liquibase actually uses (4.31.1, the version platform 2.7.4 ships):

```
PR head:  ChangeLogParseException: Error parsing line 777 column 45 of liquibase.xml:
          The content of elements must consist of well-formed character data or markup.
main:     PARSED OK: 39 changesets
```

Merged as-is, module startup fails on every deployment when OpenMRS runs this changelog, so no site gets the `previous_queue_entry` column and the code that now reads it has nothing to read. The build can't catch it: nothing in the build parses the changelog, and all 104 tests pass on the broken file because the test schema comes from hbm2ddl on H2.

`!=` is valid on MySQL, MariaDB and Postgres and keeps the plain-text style of the other `<sql>` blocks in this file. `&lt;&gt;` or a CDATA wrapper would work too.

```suggestion
                    on prev.queue_entry_id != curr2.queue_entry_id
```

The SQL itself holds up once it parses, so the escaping is the only change needed here. I ran both changesets on MySQL 8.0, MariaDB 10.11 and Postgres 14 against a `queue_entry` fixture covering a plain transition pair, two ambiguous candidates, a row whose `queue_id` equals its own `queue_coming_from` with `ended_at = started_at`, a pair with both visits null, and a voided predecessor. Each linked exactly the rows it should and left the rest null.

api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
```diff
@@ -295,6 +300,11 @@ private static Date roundToSecond(Date date) {
 	@Override
 	@Transactional(readOnly = true)
 	public QueueEntry getPreviousQueueEntry(@NotNull QueueEntry queueEntry) {
+		return queueEntry.getPreviousQueueEntry();
```
dkayiwa: This needs a voided check before merge. Reading the column straight back drops the `voided = false` filter the old criteria lookup applied through `AbstractBaseQueueDaoImpl.includeVoidedObjects`, so `undoTransition` will now act on a voided predecessor.

Merged as-is, an undo whose predecessor has since been voided (`DELETE /ws/rest/v1/queue-entry/{uuid}` routes to `voidQueueEntry`) voids the successor and "re-opens" the voided predecessor by clearing its `ended_at`. Nothing is thrown, and because the list queries exclude voided entries the patient ends up with no visible queue entry at all. I ran the same integration test on both branches (transition entry 3, void entry 3, then undo the new entry):

```
main:     undoTransition THREW IllegalArgumentException: Specified queue entry does not have a previous queue entry
          predecessor(3): voided=true  endedAt=2026-07-29 14:25:58.0    successor(11): voided=false endedAt=null
PR head:  undoTransition RETURNED id=3 voided=true endedAt=null
          predecessor(3): voided=true  endedAt=null                     successor(11): voided=true  endedAt=null
```

```suggestion
		QueueEntry previous = queueEntry.getPreviousQueueEntry();
		if (previous == null || previous.getVoided()) {
			return null;
		}
		return previous;
```

That fixes `undoTransition` but not the serialized field. With the `@PropertyGetter` gone, `BaseDelegatingResource.getProperty` falls through to `PropertyUtils.getProperty(delegate, "previousQueueEntry")`, so the REST payload reads the association directly and will surface voided predecessors it used to filter out. I would put the getter back and have it call this method, so one rule covers both paths. It costs nothing now that the method is a field read instead of a query, and the entity-level `@BatchSize` still batches the proxy initialization.

dkayiwa:
@claude review

api/src/main/java/org/openmrs/module/queue/api/QueueEntryService.java
```diff
@@ -62,13 +62,14 @@ public interface QueueEntryService {
 	List<QueueEntry> getOverlappingQueueEntries(Patient patient, Queue queue, Date startedAt, Date endedAt);
 	
 	/**
-	 * Given a specified queue entry Q, return its previous queue entry P, where P has same patient and
-	 * visit as Q, and P.endedAt time is same as Q.startedAt time, and P.queue is same as
-	 * Q.queueComingFrom
+	 * Given a specified queue entry Q, return the previous queue entry recorded against it. This is set
+	 * when Q is created, either directly by a transition or, for a queue entry saved with a
+	 * queueComingFrom, by looking up the entry P with the same patient and visit as Q, whose endedAt
```
NethmiRodrigo: Because Q can have no visit, and in that case any visit is matched
```suggestion
	 * queueComingFrom, by looking up the entry P with the same patient and, when Q has a visit, the same visit as Q, whose endedAt
```

api/src/main/resources/liquibase.xml
```diff
@@ -749,4 +749,67 @@
...
+            <not><columnExists tableName="queue_entry" columnName="previous_queue_entry"/></not>
+        </preConditions>
+        <comment>
+            Add column previous_queue_entry to queue entry table
+        </comment>
+        <addColumn tableName="queue_entry">
+            <column name="previous_queue_entry" type="int"/>
+        </addColumn>
+        <addForeignKeyConstraint baseColumnNames="previous_queue_entry" baseTableName="queue_entry" constraintName="queue_entry_previous_queue_entry_fk" onDelete="SET NULL" onUpdate="NO ACTION" referencedColumnNames="queue_entry_id" referencedTableName="queue_entry"/>
+    </changeSet>
```
NethmiRodrigo, line 823: Lets separate these two into its own changesets
```suggestion
    </changeSet>

     <changeSet id="add_previous_queue_entry_fk_20260613" author="ujjawalprabhat">
         <preConditions onFail="MARK_RAN">
             <not><foreignKeyConstraintExists foreignKeyTableName="queue_entry" foreignKeyName="queue_entry_previous_queue_entry_fk"/></not>
         </preConditions>
         <addForeignKeyConstraint baseColumnNames="previous_queue_entry" baseTableName="queue_entry" constraintName="queue_entry_previous_queue_entry_fk" onDelete="SET NULL" onUpdate="NO ACTION" referencedColumnNames="queue_entry_id" referencedTableName="queue_entry"/>
     </changeSet>
```

api/src/test/java/org/openmrs/module/queue/api/QueueEntryServiceTest.java
```diff
@@ -132,6 +133,64 @@ public void shouldCreateNewQueueEntryRecord() {
...
+		newEntry.setPriority(new Concept());
+		newEntry.setStartedAt(DateUtils.truncate(new Date(), Calendar.SECOND));
+		newEntry.setQueueComingFrom(new Queue());
+		
+		when(dao.getQueueEntries(any())).thenReturn(Collections.singletonList(prevEntry));
+		when(dao.createOrUpdate(any())).thenAnswer(invocation -> invocation.getArgument(0));
+		
+		QueueEntry result = queueEntryService.saveQueueEntry(newEntry);
+		
+		assertThat(result.getPreviousQueueEntry(), equalTo(prevEntry));
```
NethmiRodrigo, line 155: We should assert what the lookup queried here, and not just the result
```suggestion
		assertThat(result.getPreviousQueueEntry(), equalTo(prevEntry));
		verify(dao).getQueueEntries(queueEntrySearchCriteriaArgumentCaptor.capture());
		QueueEntrySearchCriteria criteria = queueEntrySearchCriteriaArgumentCaptor.getValue();
		assertThat(criteria.getPatient(), equalTo(newEntry.getPatient()));
		assertThat(criteria.getVisit(), equalTo(newEntry.getVisit()));
        assertThat(criteria.getEndedOn(), equalTo(newEntry.getStartedAt()));
        assertThat(criteria.getQueues(), equalTo(Collections.singletonList(newEntry.getQueueComingFrom())));
```

integration-tests/src/test/java/org/openmrs/module/queue/api/QueueEntryServiceTest.java
```diff
@@ -117,4 +118,70 @@ public void transitionQueueEntryShouldEndInitialIfNewIsNotDuplicate() {
...
+		QueueEntry entry3 = queueEntryService.getQueueEntryById(3).get();
+		QueueEntryTransition transition = new QueueEntryTransition();
+		transition.setQueueEntryToTransition(entry3);
+		transition.setTransitionDate(new Date());
+		QueueEntry newEntry = queueEntryService.transitionQueueEntry(transition);
+		assertThat(queueEntryService.getQueueEntryById(entry3.getQueueEntryId()).get().getEndedAt(), is(notNullValue()));
+		
+		QueueEntry reopened = queueEntryService.undoTransition(newEntry);
+		assertThat(reopened.getQueueEntryId(), is(entry3.getQueueEntryId()));
+		assertThat(newEntry.getPreviousQueueEntry(), is(nullValue()));
```
NethmiRodrigo: This reads the field undoTransition just set in the same session, so it passes even if the clear never reaches the database
```suggestion
		Context.flushSession();
		Context.clearSession();
		assertThat(queueEntryService.getQueueEntryById(newEntry.getQueueEntryId()).get().getPreviousQueueEntry(), is(nullValue()));
```

api/src/main/java/org/openmrs/module/queue/api/QueueEntryService.java
```diff
@@ -62,13 +62,14 @@ public interface QueueEntryService {
...
+	 * time is the same as Q.startedAt time, and whose queue is Q.queueComingFrom
 	 *
 	 * @param queueEntry
-	 * @return the previous queue entry, null otherwise.
-	 * @throws IllegalStateException if multiple previous queue entries are identified
+	 * @return the previous queue entry, null if there is none, if it could not be identified
+	 *         unambiguously when the queue entry was created, or if it has since been voided.
```
NethmiRodrigo: This comment seems overly complex and hard to read
```suggestion
	 * @return the previous queue entry recorded against this entry. Returns null when no predecessor
       *         was recorded at creation (no match, or more than one match), or when the recorded
       *         predecessor is voided.
```

NethmiRodrigo (changes requested):

PR #119, opened 2026-07-19
O3-5765: Automatically clear queue entries on a schedule

UjjawalPrabhat:
## Summary

Adds `AutoCloseQueueEntryTask`. Each minute it ends active entries whose `startedAt` is on or before the most recent occurrence of the configured time of day. So a run missed at that time is caught up by the next one, and entries started since then stay in the queue.

Both this and the existing visit-close task are now registered with core's scheduler as `TaskDefinitions` from `QueueModuleActivator`, replacing the module's own `QueueTaskExecutor` and `QueueTimerTask`. They appear on the Manage Scheduler page, where the interval can be changed or a task stopped.

Two new global properties:

`queue.autoCloseQueueEntriesAtTime` - `HH:mm`, blank by default, which disables clearing. An outpatient clinic would set 23:59.
`queue.autoCloseQueueEntriesForQueues` - comma-separated queue uuids; blank means all queues.

Blank by default so nothing changes for an existing deployment until an implementer opts in ([O3-2443](https://openmrs.atlassian.net/browse/O3-2443)).

## Related Issue
[O3-5765](https://openmrs.atlassian.net/browse/O3-5765)

[O3-2443]: https://openmrs.atlassian.net/browse/O3-2443?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[O3-5765]: https://openmrs.atlassian.net/browse/O3-5765?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/src/main/java/org/openmrs/module/queue/tasks/AutoCloseQueueEntryTask.java
```diff
@@ -0,0 +1,168 @@
...
+	/**
+	 * @return the queues whose entries should be cleared, or null to clear entries in all queues
+	 */
+	protected List<Queue> getQueuesToClear() {
+		String configuredQueues = getServices().getAdministrationService()
+		        .getGlobalProperty(AUTO_CLOSE_QUEUE_ENTRIES_FOR_QUEUES);
+		if (StringUtils.isBlank(configuredQueues)) {
+			return null;
+		}
+		return getServices().getQueues(configuredQueues.split(","));
```
dkayiwa: One bad UUID in `queue.autoCloseQueueEntriesForQueues` (a typo, or a queue that has since been deleted) turns this feature off with no trace at all. `getQueue()` throws `IllegalArgumentException`, the exception escapes `run()` on every tick once the close time has passed, and nothing ever surfaces it: `QueueTimerTask`'s catch can't see it because the task runs on a separate thread, and `Daemon.runInDaemonThreadWithoutResult()` submits to a plain executor and returns a `Future` that gets discarded, so the exception is captured into a `Future` nobody reads and is never logged anywhere (I checked this against openmrs-core 2.7.4, `Daemon.runInDaemonThreadInternal` and `OpenmrsThreadPoolHolder`, which is what this module builds on). If merged as-is, a misconfigured queue list means the queues silently never clear, and the logs contain nothing to debug it with.

I'd fix this before merge. Two things seem worth doing: catch and log (error level) around the body of `run()`, which also covers a transient `getQueueEntries()` failure, and resolve the refs here one at a time so a single bad UUID logs a warning naming the global property instead of aborting the whole sweep. A test that `run()` survives `getQueuesToClear()` throwing would pin the behavior.

api/src/main/java/org/openmrs/module/queue/tasks/AutoCloseQueueEntryTask.java
```diff
@@ -0,0 +1,168 @@
...
+			log.debug("There are {} queue entries to auto-close", queueEntries.size());
+			for (QueueEntry queueEntry : queueEntries) {
+				try {
+					queueEntry.setEndedAt(now);
+					saveQueueEntry(queueEntry);
+					log.info("Queue entry auto-closed on schedule: {}", queueEntry.getQueueEntryId());
+				}
+				catch (Exception e) {
+					log.warn("Unable to auto-close queue entry {}", queueEntry.getQueueEntryId(), e);
+				}
```
dkayiwa: This catch doesn't skip the entry the way it appears to. The entries from `getQueueEntries()` are managed by the daemon thread's Hibernate session (one session spans this whole `run()`), so `setEndedAt(now)` has already dirtied the entity before validation runs. Validation fires in `RequiredDataAdvice`, which core wires as a preInterceptor ahead of the transaction interceptor, meaning a `ValidationException` is thrown outside any transaction: nothing rolls back, nothing clears the session, and the entity stays dirty. When a later entry in this loop saves successfully, that commit flushes the whole session, and the rejected entry's `endedAt` goes to the database with it, validation bypassed.

The case that will actually hit this: an entry whose visit was stopped in the last minute or so, which `AutoCloseVisitQueueEntryTask` hasn't closed yet. Validation rejects `endedAt = now` for being after the visit stop date, this warn says the entry was skipped, and the next successful save in the batch persists `endedAt = now` anyway. A queue entry that ends after its visit ended is exactly the state `QueueEntryValidator` (and the guard in `QueueEntryDaoImpl.updateIfUnmodified`) exist to prevent. If the failing entry is last in the batch there is no later flush and the change is discarded instead, so the outcome even depends on batch position. I traced this through core 2.7.4 sources (`HibernateContextDAO.openSession`, the `serviceInterceptors` list in `applicationContext-service.xml`) rather than reproducing it live, so tell me if I've missed a link in the chain, but each piece is there in the source.

Evicting on failure makes the skip real, and I'd do that before this merges since it's a one-liner:

```suggestion
				catch (Exception e) {
					log.warn("Unable to auto-close queue entry {}", queueEntry.getQueueEntryId(), e);
					// the failed entity is still dirty in the session; evict it so a later flush cannot persist the rejected change
					Context.evictFromSession(queueEntry);
				}
```

`AutoCloseVisitQueueEntryTask` has the same latent problem (its `endedAt = visitStopDatetime` can equal `startedAt`, get rejected, then leak the same way), but that predates this PR; probably a follow-up rather than something to grow this PR with.

omod/src/main/resources/config.xml
```diff
@@ -55,5 +55,15 @@
         <defaultValue></defaultValue>
         <description>The bean name of a registered component that provides an algorithm to set a queue entry sort weight when saved</description>
     </globalProperty>
+    <globalProperty>
+        <property>${project.parent.artifactId}.autoCloseQueueEntriesAtTime</property>
```
dkayiwa, line 59: Can you add these two to the Global Properties section of the README as well? It documents every property with its default and behavior, and it's also the natural place to spell out what the PR description already flags: with the `23:59` default, existing deployments begin clearing all queues at end of day as soon as they pick up this version, and implementers who don't want that need to blank the property. Not blocking, but doing it in this PR keeps that section from going stale.

api/src/test/java/org/openmrs/module/queue/tasks/AutoCloseQueueEntryTaskTest.java
```diff
@@ -0,0 +1,147 @@
...
+	
+	@Before
+	public void setup() throws Exception {
+		queueEntries.clear();
+		configuredQueues = null;
+		now = getDate("2020-01-01 23:59");
+	}
+	
+	@Test
+	public void shouldDoNothingWhenTimeIsBlank() throws Exception {
```
dkayiwa, line 132: Worth adding a case for an unparseable value too (say `configuredTime = "nonsense"`), asserting nothing gets ended and no exception escapes: `getCloseTimeForToday()` handles that path by returning null, but nothing exercises it today. Take it or leave it.

omod/src/main/resources/config.xml
```diff
@@ -55,5 +55,15 @@
...
+        <defaultValue>23:59</defaultValue>
+        <description>Time of day (HH:mm, server local time) at which active queue entries are automatically ended each day. Leave blank to disable automatic clearing.</description>
+    </globalProperty>
+    <globalProperty>
+        <property>${project.parent.artifactId}.autoCloseQueueEntriesForQueues</property>
+        <defaultValue></defaultValue>
+        <description>Comma-separated list of queue uuids whose entries are automatically ended at the configured time. Leave blank to clear entries in all queues.</description>
+    </globalProperty>
```
dkayiwa, line 67: I don't think this should merge with the clear switched on for every queue by default. Between these two defaults the feature is on and unscoped: `23:59` starts the sweep and a blank queue list means all queues, so every deployment that picks up this version has every active entry force-ended at 23:59 on the first night, with one INFO line per entry and no other signal.

That is what O3-2443 removed the previous `AutoCloseQueueEntryTask` for, and its reasons still apply. It argued that queue entries now often mean "where is the patient in the clinic right now", so they may represent where inpatients currently are in the course of a multi-day visit, and that "a Queue of patients I need to follow-up with this week should not be cleared at the end of the day". A site using a queue for either of those loses that state overnight, and nothing tells them a scheduled task did it. The README note explains the default but does not help anyone who upgrades without reading it, and by then the data is gone.

The two properties are exactly the configurability O3-2443 asked for, so my suggestion is to keep them and default the time to blank, leaving the clear off until an implementer sets it. The README can say `23:59` is the value an ordinary outpatient clinic wants. That costs a fresh install one global property before it gets a scheduled clear, which I think is the better side of the trade against an existing site silently losing inpatient or follow-up queue state.

O3-5765 does say "default: end of day", so if the intent really is on out of the box then that is a maintainer call, but the two O3-2443 cases would need an answer here rather than a line in the README.

api/src/main/java/org/openmrs/module/queue/tasks/AutoCloseQueueEntryTask.java
```diff
@@ -0,0 +1,190 @@
...
+			
+			Date now = now();
+			Date closeTime = getCloseTimeForToday(configuredTime.trim(), now);
+			if (closeTime == null) {
+				return;
+			}
+			if (now.before(closeTime)) {
+				log.debug("Current time is before the configured auto-close time {}, nothing to do", configuredTime);
+				return;
+			}
```
dkayiwa: This needs fixing before merge. Together with `getCloseTimeForToday`, this guard means the sweep only ever runs between the configured time and midnight, so with the `23:59` default there is exactly one 60-second opportunity a day. If the task does not tick in that minute, the day's entries are never ended and nothing retries until 23:59 the next night: the clinic opens the following morning with yesterday's patients still queued, and nothing at the default log level says the clear was skipped. A restart or maintenance window overlapping 23:59 is enough, and so is a module start or stop near midnight, since that refreshes the Spring context and `QueueTaskExecutor` comes back with a fresh one-minute initial delay.

I ran this rather than reading it off the code. With `configuredTime = "23:59"` and an entry started `2020-01-01 09:00`, stepping `now()` across `2020-01-02` at 00:05, 06:00, 08:00, 12:00, 17:00 and 23:58 left `endedAt` null at every one of them. It was only set at `2020-01-02 23:59:30`, a full day late.

The smallest fix I can see is to resolve the most recent occurrence of the configured time instead of today's, and delete this guard:

```java
closeTime.set(Calendar.SECOND, 0);
closeTime.set(Calendar.MILLISECOND, 0);
if (closeTime.getTime().after(referenceDate)) {
	closeTime.add(Calendar.DATE, -1);
}
return closeTime.getTime();
```

`startedOnOrBefore` does the rest, and entries started after the close time still stay in the queue. The method name and its javadoc would want to lose the "ForToday". One behaviour change to know about: the first tick after an upgrade ends anything still open from before the last close time instead of waiting for tonight.

api/src/main/java/org/openmrs/module/queue/tasks/AutoCloseQueueEntryTask.java
```diff
@@ -0,0 +1,190 @@
...
+	/**
+	 * Parses the configured HH:mm time and returns the corresponding instant on the same day as the
+	 * given reference date. Returns null if the configured value cannot be parsed.
+	 */
+	protected Date getCloseTimeForToday(String configuredTime, Date referenceDate) {
+		try {
+			SimpleDateFormat format = new SimpleDateFormat(TIME_FORMAT);
+			format.setLenient(false);
+			Calendar parsed = Calendar.getInstance();
+			parsed.setTime(format.parse(configuredTime));
```
dkayiwa: `SimpleDateFormat.parse(String)` stops at the first character it cannot use and ignores the rest, so any value starting with a valid `HH:mm` is accepted whatever trails it. `11:00 PM` parses as 11:00 and the queues get cleared mid-morning, silently. I tried the other malformed values I could think of (`11 PM`, `1130`, `24:00`, `23:60`, `nonsense`) and those are all rejected properly, so it is only the case where garbage follows a valid prefix.

Not a merge blocker, but cheap to close off: parse through a `ParsePosition` and treat a null result or `position.getIndex() != configuredTime.length()` as invalid. The existing warn then covers both cases, and the `ParseException` catch goes away with it.

api/src/test/java/org/openmrs/module/queue/tasks/AutoCloseQueueEntryTaskTest.java
```diff
@@ -0,0 +1,156 @@
...
+	
+	class TestAutoCloseQueueEntryTask extends AutoCloseQueueEntryTask {
+		
+		@Override
+		protected String getConfiguredCloseTime() {
+			return configuredTime;
+		}
+		
+		@Override
+		protected List<Queue> getQueuesToClear() {
```
dkayiwa, line 68: Nothing exercises the real `getQueuesToClear()`. It is stubbed here, so the comma split, the trim, the blank-element skip, the unknown-uuid warn and the empty-list case that quietly switches the whole sweep off have no coverage in any module, and this is the piece an implementer is most likely to hand a bad value. A direct test of the real method with `getServices()` stubbed would cover it. Not a merge blocker, but I would rather see it land here than as a follow-up.

The emulation below has a smaller version of the same gap: it reads `startedOnOrBefore` and `queues` off the criteria but hardcodes `e.getEndedAt() == null` instead of deriving it from `criteria.getIsEnded()`. Drop `setIsEnded(Boolean.FALSE)` from `run()` and every test here still passes, while in production the task would start rewriting `endedAt` on entries that had already been closed.

api/src/test/java/org/openmrs/module/queue/tasks/AutoCloseQueueEntryTaskTest.java
```diff
@@ -0,0 +1,256 @@
...
+			        .filter(e -> criteria.getIsEnded() == null || criteria.getIsEnded().equals(e.getEndedAt() != null))
+			        .filter(e -> e.getStartedAt() == null || !e.getStartedAt().after(criteria.getStartedOnOrBefore()))
+			        .filter(e -> criteria.getQueues() == null || criteria.getQueues().contains(e.getQueue()))
+			        .collect(Collectors.toList());
+		}
+		
+		@Override
+		protected void saveQueueEntry(QueueEntry queueEntry) {
+			// Do nothing
+		}
```
dkayiwa, line 110: I could not find a test that reaches the evict-and-continue branch in `closeQueueEntry`, and this stub is the reason: nothing can make a save fail, so that branch never runs. It leaves the fix for the flush problem as the one behaviour here with no test behind it, and two identical `Context.evictFromSession` calls in adjacent catch blocks are just the sort of thing a later tidy-up collapses or drops.

Throwing from this stub does not fill the gap by itself. I tried it: `Context.evictFromSession` throws an `APIException` outside a service context, that throw escapes into `run()`'s catch, and the rest of the batch is skipped, so the test fails for a reason that has nothing to do with the code under test. `mockStatic(Context.class)` is no way out in the api module either, where it dies with "Cannot instrument class org.openmrs.api.context.Context because it or one of its supertypes could not be initialized" (Context's static initialiser reads runtime properties while log4j configures itself, which the omod tests survive and the api tests do not).

A seam next to the ones already on the class is enough:

```java
protected void evictFromSession(QueueEntry queueEntry) {
	Context.evictFromSession(queueEntry);
}
```

A test can then make `saveQueueEntry` throw for one entry and assert that the rest of the batch is still ended and that the failed one was evicted, which is the part that breaks if the evict calls ever go away. The stronger property, that the rejected `endedAt` never reaches the database, would need a context-sensitive test over in `integration-tests`, and I would take the cheap version here rather than grow this PR for it.

Nothing in this blocks the merge.

dkayiwa (commented):
The PR description still describes an earlier version of this. It says the default is `23:59`, warns that existing deployments will begin clearing all queues at end of day once deployed, and describes the task as acting once the clock passes the configured time. None of that holds any more: the property defaults to blank, and the task works back from the most recent occurrence of the configured time. Could you refresh it? It is the first thing anyone coming to this PR reads, and as written it advertises the behaviour that got changed.

api/src/main/java/org/openmrs/module/queue/tasks/AutoCloseQueueEntryTask.java
```diff
@@ -0,0 +1,203 @@
...
+			
+			QueueEntrySearchCriteria criteria = new QueueEntrySearchCriteria();
+			criteria.setIsEnded(false);
+			criteria.setStartedOnOrBefore(closeTime);
+			criteria.setQueues(queues);
+			
+			List<QueueEntry> queueEntries = getQueueEntries(criteria);
+			log.debug("There are {} queue entries to auto-close", queueEntries.size());
+			for (QueueEntry queueEntry : queueEntries) {
+				closeQueueEntry(queueEntry, now);
```
dkayiwa: Is `now` the value you want for `endedAt` here, or the close time the sweep is working from? On a normal night the two are within a minute of each other, but on the catch-up run the last round added they can be hours apart, and `endedAt` is what the wait-time metric is computed from.

I ran the PR's own missed-run scenario to size it: entry started `2020-01-01 09:00`, close time `2020-01-01 23:59`, server back up with its first tick at `2020-01-02 08:00`. `QueueUtils.computeAverageWaitTimeInMinutes` on the result returns 1380 minutes, against 899 if `endedAt` were the close time. That number is served as `averageWaitTime` from `/rest/v1/queue-entry-metric` and from `Legacy1xRestController`, so one night of downtime adds 481 minutes to the recorded wait of every patient still queued, and nothing fails to make it visible. It also fires on the first run after an implementer sets the property, which the README describes as expected, so this is the adoption path rather than a rare one.

My preference would be `closeTime`, which is also what `AutoCloseVisitQueueEntryTask` does next door (it records `visitStopDatetime`, not its own run time). One thing to handle if you go that way: an entry whose `startedAt` is exactly the close time is included by `startedOnOrBefore`, and `QueueEntryValidator` wants `endedAt` strictly after `startedAt`, so it would be rejected and re-warned every minute until the next day's close time moves past it. `transitionQueueEntry` already has the idiom for that collision (`new Date(currentStartedAt.getTime() + 1000L)`).

There is a reading where `now` is the right answer, of course: the entry really was open until 08:00 and nobody closed it. If you would rather keep it, I think the README needs a line saying `endedAt` is the run time, because "active queue entries are automatically ended each day" at the configured time reads as though it is that time that gets recorded.

api/src/main/java/org/openmrs/module/queue/tasks/AutoCloseQueueEntryTask.java
dkayiwa: Nothing reaches this catch: I deleted the whole `catch (Exception e)` block from `run()` and all 16 tests in `AutoCloseQueueEntryTaskTest` still pass. What it protects against is an exception vanishing into the discarded daemon `Future`, so if it is ever refactored away the failure is silent, which is the case for pinning it. Making `getQueueEntries` throw and asserting `run()` returns rather than propagating would cover it. Not something I would hold the merge for.

ibacher (commented):
While I realize this wasn't added by this module, it would be better to use core's task scheduling features rather than ScheduledExecutorTask. That said, if we are going to use Spring scheduling, then do not call `Context.getService()` but inject the service from Spring.

UjjawalPrabhat:
> it would be better to use core's task scheduling features rather than ScheduledExecutorTask. 

@ibacher Tried out using core's scheduling feat....have a look!!

api/src/main/java/org/openmrs/module/queue/tasks/AutoCloseVisitQueueEntryTask.java
```diff
@@ -62,7 +61,7 @@ public void run() {
 			}
 		}
 		finally {
-			currentlyExecuting = false;
+			stopExecuting();
```
jwnasambu, line 69: This PR establishes the `evictFromSession` pattern in `AutoCloseQueueEntryTask` to prevent a dirty entity from leaking into a later commit when save fails, but the same gap exists here. When `saveQueueEntry` throws, `endedAt` is already set on the entity (line 50) but the exception rolls back nothing: `ValidationException` fires outside any transaction, so the entity stays dirty in the Hibernate session. When a later entry in the loop saves successfully, that commit flushes the whole session and the rejected entry's `endedAt` reaches the database with validation bypassed.

Since this task is also modified in this PR (switched from `Runnable` to `AbstractTask`), the fix is the same seam already added next door:

```java
protected void evictFromSession(QueueEntry queueEntry) {
    Context.evictFromSession(queueEntry);
}
```

called from both catch blocks. The unit test would need a matching override stub and a test that makes save fail for one entry while succeeding for the next.

jwnasambu:
@claude review

README.md
```diff
@@ -69,6 +69,43 @@ A typical set of priorities might include:  `Normal` and `Emergency`
...
+started later in the day are left untouched.
+
+This property is blank by default, so no clearing happens until an implementer sets it. An ordinary outpatient clinic
+that wants its queues emptied at end of day would set `23:59`. Leave it blank if queue entries are used for anything
+that should survive overnight — tracking where inpatients currently are within a multi-day visit, for instance, or a
+queue of patients to follow up with over the coming week. Use `queue.autoCloseQueueEntriesForQueues` to enable
+clearing for some queues but not others.
+
+The task is registered with the scheduler as `Queue Module - Auto Close Queue Entries` and appears on the Manage
+Scheduler page, where its interval can be changed or the task stopped altogether.
```
NethmiRodrigo: A stop from that page doesn't survive a restart, core reschedules every task with start-on-startup set
```suggestion
Scheduler page, where its interval can be changed or the task stopped until the next restart. Blanking this
property is what turns the clear off for good.
```

api/src/main/java/org/openmrs/module/queue/QueueModuleActivator.java
```diff
@@ -9,27 +9,67 @@
...
+			return;
+		}
+		TaskDefinition taskDefinition = new TaskDefinition();
+		taskDefinition.setName(name);
+		taskDefinition.setDescription(description);
+		taskDefinition.setTaskClass(taskClass.getName());
+		taskDefinition.setStartTime(new Date());
+		taskDefinition.setRepeatInterval(REPEAT_INTERVAL_SECONDS);
+		taskDefinition.setStartOnStartup(true);
+		try {
```
NethmiRodrigo: So a failure in the lookup can't abort module startup and skip the second task
```suggestion
		try {
			SchedulerService schedulerService = Context.getSchedulerService();
			if (schedulerService.getTaskByName(name) != null) {
				log.debug("Scheduled task {} is registered already", name);
				return;
			}
			TaskDefinition taskDefinition = new TaskDefinition();
			taskDefinition.setName(name);
			taskDefinition.setDescription(description);
			taskDefinition.setTaskClass(taskClass.getName());
			taskDefinition.setStartTime(new Date());
			taskDefinition.setRepeatInterval(REPEAT_INTERVAL_SECONDS);
			taskDefinition.setStartOnStartup(true);
```

NethmiRodrigo (commented):
Thanks @UjjawalPrabhat! A few suggestions -

api/src/main/java/org/openmrs/module/queue/tasks/AutoCloseQueueEntryTask.java
```diff
@@ -0,0 +1,206 @@
...
+	
+	private void closeQueueEntry(QueueEntry queueEntry, Date closeTime) {
+		try {
+			Date endedAt = closeTime;
+			Date startedAt = queueEntry.getStartedAt();
+			if (startedAt != null && !endedAt.after(startedAt)) {
+				endedAt = new Date(startedAt.getTime() + 1000L);
+			}
+			queueEntry.setEndedAt(endedAt);
+			saveQueueEntry(queueEntry);
```
NethmiRodrigo: The query at the top of `execute()` loads every matching entry as a snapshot, and this save writes the snapshot's fields back over the row. So if an entry gets transitioned in the gap between the query and its turn in the loop, the save overwrites the newer row and its `endedAt` becomes the close time instead of the transition time, with no error anywhere. `transitionQueueEntry` handles this exact gap by [reloading the entry and saving through `updateIfUnmodified`](https://github.com/openmrs/openmrs-module-queue/blob/590b1a67fb25f4a35a979a4ceed27c741be15daa/api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java#L110-L140). Lets add a service method that does the same, reload the entry, skip it if it's already ended, save through `updateIfUnmodified`, and call it from this loop. The visit task can use it too.

api/src/main/java/org/openmrs/module/queue/tasks/AutoCloseQueueEntryTask.java
```diff
@@ -0,0 +1,206 @@
...
+				log.debug("No queues configured for auto-close, nothing to do");
+				return;
+			}
+			
+			QueueEntrySearchCriteria criteria = new QueueEntrySearchCriteria();
+			criteria.setIsEnded(false);
+			criteria.setStartedOnOrBefore(closeTime);
+			criteria.setQueues(queues);
+			
+			List<QueueEntry> queueEntries = getQueueEntries(criteria);
```
NethmiRodrigo, line 83: The first run after an implementer sets the property can load years of never-ended entries into one session, and every save's duplicate check then dirty-checks all of them again. Lets flush and clear the session every few hundred entries (behind a protected seam like `evictFromSession`) so that first sweep stays linear.

api/src/main/java/org/openmrs/module/queue/tasks/AutoCloseQueueEntryTask.java
```diff
@@ -0,0 +1,206 @@
...
+			
+			Date now = now();
+			Date closeTime = getMostRecentCloseTime(configuredTime.trim(), now);
+			if (closeTime == null) {
+				return;
+			}
+			
+			List<Queue> queues = getQueuesToClear();
+			if (queues != null && queues.isEmpty()) {
+				log.debug("No queues configured for auto-close, nothing to do");
```
NethmiRodrigo: This branch is only reached when uuids were configured but none of them resolved
```suggestion
				log.debug("None of the queues configured for auto-close could be resolved, nothing to do");
```

api/src/main/java/org/openmrs/module/queue/tasks/AutoCloseQueueEntryTask.java
```diff
@@ -0,0 +1,206 @@
...
+		}
+	}
+	
+	private void closeQueueEntry(QueueEntry queueEntry, Date closeTime) {
+		try {
+			Date endedAt = closeTime;
+			Date startedAt = queueEntry.getStartedAt();
+			if (startedAt != null && !endedAt.after(startedAt)) {
+				endedAt = new Date(startedAt.getTime() + 1000L);
+			}
```
NethmiRodrigo, line 109: Nit:
Since `startedOnOrBefore` is inclusive, an entry started exactly at the close time gets swept too, and the validator wants `endedAt` strictly after `startedAt` - lets add a comment explaining it
```suggestion
			if (startedAt != null && !endedAt.after(startedAt)) {
				// QueueEntryValidator requires endedAt to be strictly after startedAt
				endedAt = new Date(startedAt.getTime() + 1000L);
			}
```

api/src/main/java/org/openmrs/module/queue/tasks/AutoCloseQueueEntryTask.java
```diff
@@ -0,0 +1,206 @@
...
+
+/**
+ * This ends all active queue entries in the configured queues once per day at a configured time of
+ * day. The set of queues to clear and the time to clear them are both controlled by global
+ * properties, and nothing is cleared unless a time has been configured. Each run works from the
+ * most recent occurrence of that time rather than from the current tick, so patients added since
+ * then are left in the queue and a run missed at the configured time is caught up by the next one.
+ */
+@Slf4j
+public class AutoCloseQueueEntryTask extends AbstractTask {
```
NethmiRodrigo, line 40: Not part of your change, but the module already has a method that does this: [`QueueEntryService.closeActiveQueueEntries()`](https://github.com/openmrs/openmrs-module-queue/blob/590b1a67fb25f4a35a979a4ceed27c741be15daa/api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java#L253-L259) ends every active entry too, but with today's date instead of a cutoff, with no queue filter, and through the DAO directly so the validator never runs. Nothing calls it, here or in any other repo. Now that this task is the real implementation, lets delete that method in this PR so nobody reaches for the wrong one later.

api/src/test/java/org/openmrs/module/queue/tasks/AutoCloseQueueEntryTaskTest.java
```diff
@@ -0,0 +1,323 @@
...
+	
+	@Test
+	public void shouldEndEntriesStartedAtTheCloseTimeOneSecondLater() throws Exception {
+		configuredTime = "18:00";
+		now = getDate("2020-01-01 18:30");
+		QueueEntry queueEntry = queueEntryStartedAt("2020-01-01 18:00", null);
+		
+		new TestAutoCloseQueueEntryTask().execute();
+		assertThat(queueEntry.getEndedAt(), equalTo(new Date(getDate("2020-01-01 18:00").getTime() + 1000L)));
+	}
```
NethmiRodrigo, line 195: The scheduler reuses one task instance across ticks, so lets cover a second `execute()` on the same instance to pin the `isExecuting` reset
```suggestion
	}
	
	@Test
	public void shouldClearEntriesOnALaterRunOfTheSameTaskInstance() throws Exception {
		configuredTime = "23:59";
		TestAutoCloseQueueEntryTask task = new TestAutoCloseQueueEntryTask();
		task.execute();
		
		QueueEntry queueEntry = queueEntryStartedAt("2020-01-02 09:00", null);
		now = getDate("2020-01-02 23:59");
		task.execute();
		assertThat(queueEntry.getEndedAt(), equalTo(now));
	}
```

integration-tests/src/test/java/org/openmrs/module/queue/QueueModuleActivatorTest.java
```diff
@@ -0,0 +1,71 @@
...
+		new QueueModuleActivator().started();
+	}
+	
+	@Test
+	public void shouldRegisterAndStartBothTasks() {
+		assertThat(taskDefinition(AUTO_CLOSE_VISIT_QUEUE_ENTRY_TASK).getTaskClass(),
+		    equalTo(AutoCloseVisitQueueEntryTask.class.getName()));
+		assertThat(taskDefinition(AUTO_CLOSE_QUEUE_ENTRY_TASK).getTaskClass(),
+		    equalTo(AutoCloseQueueEntryTask.class.getName()));
+		assertThat(taskDefinition(AUTO_CLOSE_QUEUE_ENTRY_TASK).getStarted(), equalTo(true));
```
NethmiRodrigo, line 47: So a units mix-up in the interval or a lost start-on-startup can't slip through
```suggestion
		assertThat(taskDefinition(AUTO_CLOSE_QUEUE_ENTRY_TASK).getStarted(), equalTo(true));
		assertThat(taskDefinition(AUTO_CLOSE_QUEUE_ENTRY_TASK).getRepeatInterval(), equalTo(60L));
		assertThat(taskDefinition(AUTO_CLOSE_QUEUE_ENTRY_TASK).getStartOnStartup(), equalTo(true));
```

NethmiRodrigo:
@wikumChamith can you review this?

api/src/main/java/org/openmrs/module/queue/api/dao/impl/QueueEntryDaoImpl.java
wikumChamith: This JPQL update skips `AuditableInterceptor`, so auto-closed entries no longer get `date_changed` / `changed_by` (the old `saveQueueEntry` path set them). It also means the `dateChanged` guard can't catch a concurrent transition or close, since those never bump `date_changed` either.

I think we should  set `qe.dateChanged = :now, qe.changedBy = :changedBy` here and add `AND qe.endedAt IS NULL` to the WHERE.

ibacher: We could also re-work `AuditableInterceptor` and co to work at a lower level, but yeah. I would also say, as a rule, it is substantially better to use the JPA Criteria API to build dynamic queries than this king string munging.

api/src/main/java/org/openmrs/module/queue/api/impl/QueueEntryServiceImpl.java
```diff
@@ -250,12 +251,56 @@ public String generateVisitQueueNumber(Location location, Queue queue, Visit vis
 		return queueNumber;
 	}
 	
+	/**
+	 * @see QueueEntryService#closeQueueEntry(QueueEntry, Date)
+	 */
 	@Override
-	public void closeActiveQueueEntries() {
```
ibacher: It's not great to be deleting public API even if we no longer have a use for it. It's public and therefore someone might be relying on it.

api/src/main/java/org/openmrs/module/queue/tasks/AutoCloseQueueEntryTask.java
```diff
@@ -0,0 +1,228 @@
...
+	}
+	
+	private void closeQueueEntry(QueueEntry queueEntry, Date closeTime) {
+		try {
+			Date endedAt = closeTime;
+			Date startedAt = queueEntry.getStartedAt();
+			if (startedAt != null && !endedAt.after(startedAt)) {
+				// startedOnOrBefore is inclusive, so an entry started exactly at the close time is swept
+				// too, and QueueEntryDaoImpl.updateIfUnmodified rejects an endedAt that is not strictly
+				// after startedAt
```
ibacher: This is the main part we need to keep. It doesn't explain why we're adding 1000 here. Would 1 work as well? Why not?

```suggestion
				// QueueEntryDaoImpl.updateIfUnmodified rejects an endedAt that is not strictly
				// after startedAt
```

api/src/main/java/org/openmrs/module/queue/tasks/AutoCloseQueueEntryTask.java
```diff
@@ -0,0 +1,228 @@
...
+		}
+	}
+	
+	/**
+	 * Parses the configured HH:mm time and returns the most recent instant at which that time of day
+	 * occurred: today's occurrence if it has already passed, otherwise yesterday's. Returns null if the
+	 * configured value cannot be parsed.
+	 */
+	protected Date getMostRecentCloseTime(String configuredTime, Date referenceDate) {
+		SimpleDateFormat format = new SimpleDateFormat(TIME_FORMAT);
```
ibacher: We should not be using `SimpleDateFormat` in new code. `DateTimeFormatter` is superior, especially for cases like this where there's a fixed, static format because we can create a single `DateTimeFormatter` instance, since it, unlike `SimpleDateFormat` is thread-safe.

README.md
```diff
@@ -69,6 +69,44 @@ A typical set of priorities might include:  `Normal` and `Emergency`
...
+
+The task is registered with the scheduler as `Queue Module - Auto Close Queue Entries` and appears on the Manage
+Scheduler page, where its interval can be changed or the task stopped until the next restart. Blanking this
+property is what turns the clear off for good.
+
+The task works from the most recent occurrence of the configured time rather than from the moment it happens to run,
+so if it does not get a chance to run at that time (a restart or a maintenance window, say) the next run catches up
+instead of leaving the queues uncleared until the following day. One consequence: the first run after this property
+is set ends anything still open from before the most recent occurrence of the configured time, rather than waiting
+for the next one.
```
ibacher: This is way too long. Shorten it down to the essentials.

README.md
```diff
@@ -69,6 +69,44 @@ A typical set of priorities might include:  `Normal` and `Emergency`
...
+
+**Required?**  False
+
+**Description:**  
+The time of day, in `HH:mm` 24-hour format and the server's local time, at which active queue entries are
+automatically ended each day. A scheduled task ends every queue entry that was still active as of this time; entries
+started later in the day are left untouched.
+
+This property is blank by default, so no clearing happens until an implementer sets it. An ordinary outpatient clinic
+that wants its queues emptied at end of day would set `23:59`. Leave it blank if queue entries are used for anything
```
ibacher: Why would they want 23:59 and not 00:00?

README.md
```diff
@@ -69,6 +69,44 @@ A typical set of priorities might include:  `Normal` and `Emergency`
...
+for the next one.
+
+#### queue.autoCloseQueueEntriesForQueues
+
+**Default Value:**  None (empty)
+
+**Required?**  False
+
+**Description:**  
+A comma-separated list of queue uuids whose entries are automatically ended at the time configured by
```
ibacher: Do we have a better way to identify queues than UUIDs? Generally we want user-facing configurations to be user-understandable. UUIDs optimize for computers and should not be used if avoidable.

ibacher (commented):
Thanks @UjjawalPrabhat! A random smattering of curmudgeonly comments.

NethmiRodrigo:
@UjjawalPrabhat let me know if you'll be addressing the review, else I'll pick it up :D

NethmiRodrigo:
Ping @wikumChamith can you re-review this?

integration-tests/src/test/java/org/openmrs/module/queue/tasks/AutoCloseQueueEntryTaskIntegrationTest.java
```diff
@@ -0,0 +1,223 @@
...
+	
+	@Test
+	public void shouldNotEndAQueueEntryAfterItsVisitStoppedWhileTheTaskWasRunning() throws Exception {
+		Date visitStoppedAt = DateUtils.truncate(DateUtils.addHours(new Date(), -2), Calendar.SECOND);
+		Visit visit = Context.getVisitService().getVisit(101);
+		Integer queueEntryId = activeQueueEntryStartedHoursAgo(3, visit);
+		Context.flushSession();
+		Context.clearSession();
+		// the copy a task holds, with its visit loaded into the session while the visit was still open
+		QueueEntry loadedByTheTask = queueEntryService.getQueueEntryById(queueEntryId).get();
```
dkayiwa: Worth one more line here so this test keeps covering the visit refresh once it lands on main. O3-5763 (#122) made `QueueEntry.visit` lazy, so on main `getQueueEntryById` no longer loads the visit. It is first read inside `closeQueueEntry`, after the UPDATE below, and comes back stopped whether or not `QueueEntryDaoImpl.refresh` refreshes it. I merged this branch into main and deleted the `session.refresh(queueEntry.getVisit())` block, and all five tests in this class still pass. On this branch alone the same deletion fails this test (the entry ends at the close time, an hour after the visit stopped), so it only stops guarding the refresh after the merge.

Reading the visit before the UPDATE puts the stale copy back in the session. With this, the same deletion fails the test on the merged code again, and the test passes with the refresh in place on both main and this branch:

```suggestion
		QueueEntry loadedByTheTask = queueEntryService.getQueueEntryById(queueEntryId).get();
		// QueueEntry.visit is lazy, so read it here or it is first loaded after the update below
		assertThat(loadedByTheTask.getVisit().getStopDatetime(), nullValue());
```

The code itself is fine, so this doesn't block the merge.

integration-tests/src/test/java/org/openmrs/module/queue/QueueModuleActivatorTest.java
```diff
@@ -0,0 +1,73 @@
...
+public class QueueModuleActivatorTest extends BaseModuleContextSensitiveTest {
+	
+	private static final String AUTO_CLOSE_VISIT_QUEUE_ENTRY_TASK = "Queue Module - Auto Close Visit Queue Entries";
+	
+	private static final String AUTO_CLOSE_QUEUE_ENTRY_TASK = "Queue Module - Auto Close Queue Entries";
+	
+	@Before
+	public void setup() {
+		new QueueModuleActivator().started();
+	}
```
dkayiwa: These tests leave both tasks scheduled for the rest of the test run. `started()` puts the two definitions on the real `TimerSchedulerServiceImpl` and nothing here stops them, so from a minute after each test they fire every minute on a daemon thread, running both tasks against the database every later integration test shares. I checked with a throwaway test that calls `started()` and then waits 75 seconds: both definitions pick up a `lastExecutionTime` at the one-minute mark. Nothing trips over it today, since every test in the module rolls back, but an untransacted test that runs a minute or more after this class would share its committed rows with two scheduled tasks nothing in that test set up, which is a miserable flake to chase.

Stopping them in an `@After` keeps it contained. With this in place the throwaway test sees no execution at all, and the three tests here still pass:

```suggestion
	@Before
	public void setup() {
		new QueueModuleActivator().started();
	}
	
	@After
	public void stopTasks() throws SchedulerException {
		// started() schedules real timers, which would otherwise go on running both tasks every minute
		// on a background thread for the rest of the test run
		Context.getSchedulerService().shutdownTask(taskDefinition(AUTO_CLOSE_VISIT_QUEUE_ENTRY_TASK));
		Context.getSchedulerService().shutdownTask(taskDefinition(AUTO_CLOSE_QUEUE_ENTRY_TASK));
	}
```

It needs `import org.junit.After;` too. Not something I'd hold the merge for.

PR #120, opened 2026-07-25
O3-5760: Return null rather than NaN for the average wait time

UjjawalPrabhat:
## Summary
Came up while reviewing [openmrs-esm-patient-management#2615](https://github.com/openmrs/openmrs-esm-patient-management/pull/2615#discussion_r3606177037), where the Service Queues dashboard rendered "NaN mins".

`QueueUtils.computeAverageWaitTimeInMinutes` only guards against an empty entry list, but it averages over entries that have both a `startedAt` and an `endedAt`. A non-empty list where nothing has ended leaves the denominator at zero, so `queue-entry-metrics?metric=averageWaitTime` returns `{"averageWaitTime":"NaN"}`.

Returns `null` instead when there is no completed wait to average.

## Related Issue
[O3-5760](https://openmrs.atlassian.net/browse/O3-5760)

## Other
Related PR [here](https://github.com/openmrs/openmrs-esm-patient-management/pull/2615).

[O3-5760]: https://openmrs.atlassian.net/browse/O3-5760?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/src/test/java/org/openmrs/module/queue/utils/QueueUtilsTest.java
+++ b/api/src/test/java/org/openmrs/module/queue/utils/QueueUtilsTest.java
@@ -11,10 +11,14 @@ package org.openmrs.module.queue.utils;
 
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.nullValue;
 
+import java.util.Arrays;
 import java.util.Date;
+import java.util.List;
 
 import org.junit.Test;
+import org.openmrs.module.queue.model.QueueEntry;
 
 public class QueueUtilsTest {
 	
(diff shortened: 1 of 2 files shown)


api/src/test/java/org/openmrs/module/queue/utils/QueueUtilsTest.java
```diff
@@ -52,4 +56,31 @@ public void shouldReturnTrueIfDatesOverlap() {
...
+	
+	@Test
+	public void shouldComputeAverageWaitTimeInMinutes() {
+		// Entries that waited 24 and 48 hours average out to 36 hours
+		assertThat(QueueUtils.computeAverageWaitTimeInMinutes(entries(entry(AUG_1, AUG_2), entry(AUG_1, AUG_3))),
+		    is(2160.0));
+		
+		// Entries that have not ended yet are not part of the average
+		assertThat(QueueUtils.computeAverageWaitTimeInMinutes(entries(entry(AUG_1, AUG_2), entry(AUG_1, NULL))), is(1440.0));
+
```
NethmiRodrigo, line 70: Lets add a test case that covers a scenario where both timestamps are missing
```suggestion
		 // Entries missing a start or an end are not part of the average
             assertThat(QueueUtils.computeAverageWaitTimeInMinutes(entries(entry(AUG_1, AUG_2), entry(AUG_1, NULL))), is(1440.0));
             assertThat(QueueUtils.computeAverageWaitTimeInMinutes(entries(entry(NULL, AUG_2), entry(AUG_1, AUG_2))), is(1440.0));
```

NethmiRodrigo (commented):
Thanks @UjjawalPrabhat! Just one minor thing

NethmiRodrigo (approved):
Thanks @UjjawalPrabhat!

